PR #1, opened 2020-05-11
API Improvements and documentation

gdude2002:
This is kind of a big PR, but:

* Command system revamped, with argument parsing, signature generation and a CommandContext object.
* Additional check combinator.
* Wiki-style documentation.


docs/src/orchid/resources/wiki/components/command.md
```diff
@@ -0,0 +1,167 @@
...
+You will likely want to write commands that are only invoked based on a set of conditions.
+In order to facilitate this in a manner that allows you to easily reuse these conditions,
+commands support a [checks system](check).
+
+```kotlin
+suspend fun defaultCheck(event: MessageCreateEvent): Boolean {
+    with(event) {
+        return when {
+            message.author?.id == bot.kord.getSelf().id -> false  // Check that we didn't send this message.
+            message.author!!.isBot == true              -> false  // Check that another bot didn't send this message.
```
Akarys42, line 71: Why a safe call followed by a not-null assertion? You seem to be using both through the wiki pages, it would be nice to select one.

docs/src/orchid/resources/wiki/components/event-handler.md
```diff
@@ -0,0 +1,73 @@
+### Event Handlers
+
+Event handlers function similarly to [commands](./command), but they react to Kord events instead of command
+invocations on Discord. For a full list of Kord events, you can 
+[check out the Kord GitLab](https://gitlab.com/kordlib/kord/-/tree/master/core/src/main/kotlin/com/gitlab/kordlib/core/event)
+\- but for the purposes of this pave, we'll just be looking at a `MessageCreateEvent` - an event that is fired when
```
Akarys42, line 6: Typo
```suggestion
\- but for the purposes of this page, we'll just be looking at a `MessageCreateEvent` - an event that is fired when
```

docs/src/orchid/resources/wiki/components/command.md
gdude2002: It was still appropriate to use `!!` in some cases (or just for simplification), but I've updated most instances.

PR #2, opened 2020-05-19
Paginator and help

Akarys42:
This PR adds the paginator from Kotlin-Discord/bot#6 and the help command from Kotlin-Discord/bot#5.


Akarys42:
It isn't running any check indeed, but I don't see which checks to
actually run. The only thing I could think of is checking that the command
was invocated inside the command channel, which should be done at the bot
level, right?

>

gdude2002:
The checks defined on the Command object?

Akarys42:
There wasn't a plan about defining bot-wide checks? It would be nice to
have something like that.

gdude2002:
I didn't already have a TODO for that, no, it's more complicated than it
sounds. But we can investigate that later.

gdude2002:
We'll worry about checks later.

PR #3, opened 2020-05-22
Make help command run checks

Akarys42:
*It was easier than expected*


kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/Command.kt
```diff
@@ -180,6 +180,16 @@ class Command(val extension: Extension) {
         signature = strings.joinToString(" ")
     }
 
+    /** Run checks with the provided [MessageCreateEvent]. Return false if any failed, true otherwise. */
```
gdude2002, line 183: That should end with `**/`, but that's all. You can squash and merge once that's done.

PR #4, opened 2020-05-23
Generic checks

gdude2002:
This rewrites the checks system, so that all included checks can handle a wide range of events.

### DIFF
--- a/docs/src/orchid/resources/wiki/components/check.md
+++ b/docs/src/orchid/resources/wiki/components/check.md
@@ -105,7 +105,7 @@ I'm not your mom.
 #### Combinators
 
 If you don't want to simply require that all checks pass for a command or event handler, you can use
-a combinator check. We currently bundle one such check: `or`.
+a combinator check. We currently bundle two checks: `or` and `and`.
 
 ```kotlin
 event<MessageCreateEvent> {
--- a/docs/src/orchid/resources/wiki/components/check.md
+++ b/docs/src/orchid/resources/wiki/components/check.md
@@ -70,6 +70,38 @@ event<MessageCreateEvent> {
 
 This allows us to create far more useful checks that can operate across a variety of options.
 
+#### Generic checks
+
+In some cases, you'll want to write checks that can deal with multiple types of event. In order to facilitate
+this, we provide a `CheckUtils` module containing a set of functions that can return Kord Behaviors for a given
+event. For example, a check written to support events that concern specific members might look like this:
+
+```kotlin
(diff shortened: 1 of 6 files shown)


kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/checks/ChannelChecks.kt
```diff
@@ -2,130 +2,211 @@
...
-        with(event) {
-            return message.channel.id == channel.id
+fun inChannel(channel: ChannelBehavior): suspend (Event) -> Boolean {
+    val logger = KotlinLogging.logger {}
+
+    suspend fun inner(event: Event): Boolean {
+        val eventChannel = channelFor(event)
+
+        if (eventChannel == null) {
+            logger.debug { "Channel for event $event is null. This type of event may not be supported." }
```
Akarys42, line 26: I feel like this should be a warning instead, or at least an info, in case debug level loggings are disabled, this can save some troubleshooting.

docs/src/orchid/resources/wiki/components/check.md
```diff
@@ -70,6 +70,38 @@ event<MessageCreateEvent> {
...
+* `memberFor(event: Event): MemberBehavior?`
+* `messageFor(event: Event): MessageBehavior?`
+* `roleFor(event: Event): RoleBehavior?`
+* `userFor(event: Event): UserBehavior?`
+
+These functions will return `null` for an unsupported event type. All the checks bundled with this framework
+support generic events - if a check receives an unsupported type, that check will simply fail. These failures
+will be logged at debug level, so enable SLF4J debug logging if you're not sure - or just use a debugger,
+I'm not your mom.
+
```
Akarys42, line 104: in the next paragraph, we now also bundle the `and` check.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/checks/ChannelChecks.kt
gdude2002: This is true, but the main reason I did it is that this message really only matters to the developer. Additionally, it's going to get noisy - there are lots of _valid_ event types that will also return `null` for these function calls.

Akarys42: Ah I see the issue now, debug level seems fair.

PR #5, opened 2020-06-02
Add support for keyworded arguments and message parsing.

Akarys42:
Arguments can now be supplied in the argName:value form, no matter their position in the command. Additionally, message parsing has been added.


Akarys42:
> Is there no way for us to get a message object by ID if it's not in the cache?
The kord guys said no, because we need the channel ID, although it might be worth supporting message jump links I think. 
We also need to write about this feature in the wiki, are you still up for the task @gdude2002 ? :D

PR #7, opened 2020-07-04
Add the duration parser utility

Akarys42:
This PR adds a (sort of) standalone duration parser. 

I decided to not hook it up to the argument parser because I didn't actually need it, do you think it is a good idea to still hook it up?


ks129:
I think maybe this should allow using multiple ways to provide time unit (`m|min|minute(s)`, `s|sec|second(s)` etc.)?

Akarys42:
The magic number is just the 10 on line 33, do we really need a comment for that?

gdude2002:
Well, I wouldn't be asking if I knew what it did!

Akarys42:
`mi` and `mo` sounds good to me, it is less ambiguous this way.

PR #8, opened 2020-07-17
Add Scheduler class

Akarys42:
I think it is pretty self-explanatory, but the `Scheduler` class provides a simple interface for scheduling function calls. 

There's one thing I'm not sure about is the scope used to launch coroutines; [the docs state](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-global-scope/) that using `GlobalScope` is a bad practice, so I ended up using a custom `CoroutineScope`, but I'm unsure about what to use for its `context`, for now, I used an `EmptyContext`.

### DIFF
--- a/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/checks/utils/Scheduler.kt
+++ b/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/checks/utils/Scheduler.kt
@@ -1,17 +1,88 @@
 package com.kotlindiscord.kord.extensions.checks.utils
 
-import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.Job
-import kotlinx.coroutines.delay
-import kotlinx.coroutines.launch
+import kotlinx.coroutines.*
+import mu.KotlinLogging
 
+private val logger = KotlinLogging.logger {  }
+
+/**
+ * Class in charge of providing scheduling functions.
+ */
 class Scheduler {
     private val jobMap: MutableMap<Int, Job> = mutableMapOf()
 
-    fun <T> schedule(id: Int, delay: Long, data: T?, callback: (T?) -> Unit) {
-        jobMap[id] = GlobalScope.launch {
+    private val finishTask = CancellationException()
+
+    /**
+     * Schedule a new task.
+     *
+     * @param id The ID of the task.
+     * @param delay How long in milliseconds to wait before executing the callback.
+     * @param data Arbitrary data to be passed to the callback.
+     * @param callback Function to be executed.
+     *
+     * @return Return true on success, false otherwise.
+     */
+    fun <T> schedule(id: Int, delay: Long, data: T?, callback: (T?) -> Unit): Boolean {
+        logger.debug { "Scheduling task $id" }
+
+        if (id in jobMap) {
+            logger.warn { "Cannot schedule task, ID already in use." }
+            return false
+        }
+
+        val job = GlobalScope.launch {
             delay(delay)
             callback(data)
+        }
+
+        job.invokeOnCompletion {
+            if (it == finishTask) {
+                callback(data)
+            }
+        }
+
+        jobMap[id] = job
+        return true
+    }
+
+    /**
+     * Immediately execute the callback of the provided task.
+     *
+     * @param id ID of the targeted task.
+     */
+    fun finishJob(id: Int) {
+        logger.debug { "Finishing task $id" }
+        jobMap[id]?.cancel(finishTask)
+    }
+
+    /**
+     * Cancel the provided task.
+     *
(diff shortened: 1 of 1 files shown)


BartArys:
Using `GlobalScope` in this situation shouldn't be considered a bad practice. [This article by Roman Elizarov](https://medium.com/@elizarov/the-reason-to-avoid-globalscope-835337445abc) should help clarify why you generally don't want to use Globalscope:

> `GlobalScope.launch` creates global coroutines. It is now developer’s responsibility to keep track of their lifetime. We can “fix” an approach with `GlobalScope` by manually keeping track of the launched coroutines and waiting for their completion using join 
> ...
> but takes quite more effort, so why bother writing more code? There is hardly ever reason to use GlobalScope in an application that is based on Kotlin coroutines.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/checks/utils/Scheduler.kt
```diff
@@ -0,0 +1,90 @@
...
+     * Schedule a new task.
+     *
+     * @param id The ID of the task.
+     * @param delay How long in milliseconds to wait before executing the callback.
+     * @param data Arbitrary data to be passed to the callback.
+     * @param callback Function to be executed.
+     *
+     * @return Return true on success, false otherwise.
+     */
+    fun <T> schedule(id: Int, delay: Long, data: T?, callback: (T?) -> Unit): Boolean {
```
gdude2002, line 28: I don't think we really want users to have to specify their own ID. You could use a UUID instead, and return it - since we'd only ever return `false` here if the ID is already scheduled.

If you can think of a legitimately useful reason to be able to provide a specific ID for a job here, then you can create an overloaded method that does that.

Also, how about scheduling repeating jobs?

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/checks/utils/Scheduler.kt
```diff
@@ -0,0 +1,90 @@
...
+
+private val logger = KotlinLogging.logger {  }
+
+/**
+ * Class in charge of providing scheduling functions.
+ */
+class Scheduler {
+    private val jobMap: MutableMap<Int, Job> = mutableMapOf()
+
+    private val scope = CoroutineScope(EmptyCoroutineContext)
```
gdude2002, line 15: GlobalScope is just fine, as @BartArys has stated - but if you feel like it would be useful, keep the variable and make this an `open class` so that users can subclass it for their own needs.

I'm not sure how useful that'd be, though.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/checks/utils/Scheduler.kt
```diff
@@ -0,0 +1,90 @@
...
+     * @param callback Function to be executed.
+     *
+     * @return Return true on success, false otherwise.
+     */
+    fun <T> schedule(id: Int, delay: Long, data: T?, callback: (T?) -> Unit): Boolean {
+        logger.debug { "Scheduling task $id" }
+
+        if (id in jobMap) {
+            logger.warn { "Cannot schedule task, ID already in use." }
+            return false
```
gdude2002, line 33: If you decide not to generate IDs automatically, then this should raise an exception.

Heck, if you do decide to generate IDs automatically and allow users to optionally provide their own IDs, then it should still raise one.

Exceptional circumstances require exceptions, it should not be an "okay sure whatever" to schedule jobs with duplicate IDs

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/checks/utils/Scheduler.kt
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        val job = scope.launch {
+            delay(delay)
+            callback(data)
```
gdude2002, line 38: It might be worth making callbacks optional, for jobs where the invoker doesn't need to worry about what happens after.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/checks/utils/Scheduler.kt
ks129: Then this should return this ID, otherwise, it's unable to cancel/complete jobs.

gdude2002: Yup, that's what I said!

gdude2002:
Let's get that bread.

PR #9, opened 2020-08-11
Create base exception for parsers and throw custom exception when invalid time unit provided to parser

ks129:
Currently this throws `NullPointerException` when invalid time unit provided, but catching `NullPointerException` is too generic, so now this throw custom exception.

### DIFF
--- a/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/Exceptions.kt
+++ b/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/Exceptions.kt
@@ -88,12 +88,3 @@ class ParseException(var reason: String) : ExtensionsException() {
 
     override fun toString(): String = reason
 }
-
-/**
- * Throws when invalid time unit given to duration parser.
- *
- * @param unit Invalid unit.
- */
-class InvalidTimeUnitException(var unit: String) : ExtensionsException() {
-    override fun toString(): String = "Invalid time unit provided: $unit"
-}
--- a/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/Exceptions.kt
+++ b/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/Exceptions.kt
@@ -88,3 +88,12 @@ class ParseException(var reason: String) : ExtensionsException() {
 
     override fun toString(): String = reason
 }
+
+/**
+ * Throws when invalid time unit given to duration parser.
+ *
+ * @param unit Invalid unit.
+ */
+class InvalidTimeUnitException(var unit: String) : ExtensionsException() {
+    override fun toString(): String = "Invalid time unit provided: $unit"
+}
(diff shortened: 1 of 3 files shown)


gdude2002:
This is definitely an improvement, but would it be better to add [a custom exception](https://github.com/Kotlin-Discord/kord-extensions/blob/master/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/Exceptions.kt) for this? It would be fairly trivial, and provide users with a much better idea of what's going on.

Akarys42:
I agree with @gdude2002, we should provide a generic exception for all our (future) parsers, maybe something like a `ParsingFailedException`

ks129:
@gdude2002 @Akarys42 Is this now better?

gdude2002:
This looks fine, but [the linter is failing](https://github.com/Kotlin-Discord/kord-extensions/pull/9/checks#step:4:40).

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/parsers/DurationParser.kt
```diff
@@ -58,8 +60,13 @@ fun parseDuration(s: String): Duration {
         val r2 = buffer.splitOn { it.isDigit() }
         val unit = r2.first
         buffer = r2.second
-
-        val chronoUnit = unitMap[unit.toLowerCase()]
+        
+        val chronoUnit: ChronoUnit
+        try {
+            chronoUnit = unitMap.getValue(unit.toLowerCase())
```
Akarys42, line 66: You can do that with an Elvis operator rather than a full try except
```kotlin
val chronoUnit = unitMap[unit.toLowerCase()] ?: throw ...
```

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/Exceptions.kt
```diff
@@ -88,3 +88,12 @@ class ParseException(var reason: String) : ExtensionsException() {
 
     override fun toString(): String = reason
 }
+
+/**
+ * Throws when invalid time unit given to duration parser.
+ *
+ * @param unit Invalid unit.
+ */
+class InvalidTimeUnitException(var unit: String) : ExtensionsException() {
```
Akarys42, line 97: Like I said, that would be better if we used a common base exception for all the parsers, maybe even define them in the `parsers` folder

Akarys42:
Looks good to me, but you have some linting errors
```
empty-blocks - 5min debt
	EmptyDefaultConstructor - [BaseParserException] at /home/runner/work/kord-extensions/kord-extensions/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/parsers/Exceptions.kt:8:31
formatting - 10min debt
	NoUnusedImports - [] at /home/runner/work/kord-extensions/kord-extensions/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/parsers/DurationParser.kt:6:1
	NoSemicolons - [] at /home/runner/work/kord-extensions/kord-extensions/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/parsers/Exceptions.kt:3:61
style - 5min debt
	UnusedImports - [DurationParser.kt] at /home/runner/work/kord-extensions/kord-extensions/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/parsers/DurationParser.kt:6:1```

ks129:
@Akarys42 Now this is fixed

Akarys42 (approved):
Perfect! Thanks for your contribution!

PR #10, opened 2020-08-24
Reloadable extensions

gdude2002:
This PR updates the API a bit to allow for reloadable extensions.

This does not implement an extension that reloads other extensions - it just provides a way for extensions to be unloaded and loaded.

This PR also implements custom events that are fired when extensions are loaded and unloaded.

### DIFF
--- a/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/ExtensibleBot.kt
+++ b/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/ExtensibleBot.kt
@@ -9,14 +9,17 @@ import com.gitlab.kordlib.core.on
 import com.gitlab.kordlib.gateway.builder.PresenceBuilder
 import com.kotlindiscord.kord.extensions.commands.Command
 import com.kotlindiscord.kord.extensions.events.EventHandler
+import com.kotlindiscord.kord.extensions.events.ExtensionEvent
 import com.kotlindiscord.kord.extensions.extensions.Extension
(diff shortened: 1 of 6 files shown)


kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/ExtensibleBot.kt
```diff
@@ -192,14 +233,49 @@ open class ExtensibleBot(
...
+
+    /**
+     * Unload an installed [Extension] from this bot, by name.
+     *
+     * This function **does not** remove the extension object - it simply
+     * removes its event handlers and commands. Unloaded extensions can
+     * be loaded again by calling [ExtensibleBot.loadExtension].
+     *
+     * @param extension The name of the [Extension] to unload.
+     */
```
Akarys42, line 270: Maybe we should note that the function just returns if no matching extension is found?

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/events/EventHandler.kt
```diff
@@ -29,6 +29,11 @@ class EventHandler<T : Event>(val extension: Extension, val type: KClass<*>) {
      */
     val checkList: MutableList<suspend (T) -> Boolean> = mutableListOf()
 
+    /**
+     * @suppress
+     */
+    var job: Job? = null
```
Akarys42, line 35: Maybe we should document that? I have no idea what that is

Akarys42 (approved):
Pretty good, just two comments related to documentation

PR #11, opened 2020-09-17
Grouped commands / Additional modularity

gdude2002:
This adds a `group` function to extensions and a new `GroupCommand` class.

`GroupCommand`s are essentially _command commands_ - they're commands with their own subcommands.
Grouped commands can also contain their own command groups, because who doesn't love nesting things waaaaaaaaaaaaaay too deep.

This also updates the help extension in line with the other changes. That said, this needs testing - any takers?

---

I've also taken the opportunity to modify a bunch of stuff because extensibility to the max is quite important - now you can register custom command classes, and a bunch of classes may be subclassed and have their attributes and methods overridden.


Akarys42 (approved):
Looks good to me

PR #17, opened 2021-01-11
feat/utils Adding extension methods

Distractic:
# Adding extension methods

## Context
This PR aims to add functionalities in order to be able to handle certain objects more easily.
For messages, interacting with emojis was not easy.

### Reaction
To add a reaction to a message, it is now possible to directly use the unicode code
```kotlin
message.addReaction("🦈")
```

A String in unicode format is automatically transformed into ReactionEmoji. To do this manually you can also use the added toReaction methods
```kotlin
val react1 = "🦈".toReaction()
val react2 = emojiGuild.toReaction()
```

### User
When a command is called, or often wishes to do the processing only if the author is not null and he is not a bot. For this the following method allows to do this verification
```kotlin
val author = message.author
if(author.isNullOrBot()) return@action
val username: String = author.username //author not null automatically
```

### Member
Methods with multiple parameters instead of just one have been added. In particular to know if a member has one or more roles put in parameter
```kotlin
member.hasRole(role)
member.hasRoles(role1, role2, role3) //..
member.hasRoles(listOf(role1, role2, role3)) //..
```
Same for permissions
```kotlin
member.hasPermission(perm)
member.hasPermissions(perm1, perm2, perm3) //..
member.hasPermissions(listOf(perm1, perm2, perm3)) //..
```

### RestRequestException
When you manipulate HTTP calls, you may receive an error. For some processing you may need to know whether or not an HTTP code corresponds to the response received
```kotlin
try {
        //request
} catch (ex: RestRequestException) {
  when{
     ex.hasStatus(HttpStatusCode.Forbidden) -> null
     ex.hasStatus(HttpStatusCode.BadGateway, HttpStatusCode.RequestURITooLong) -> null
     ex.hasNotStatus(HttpStatusCode.ExpectationFailed) -> null
     ex.hasNotStatus(HttpStatusCode.UnprocessableEntity, HttpStatusCode.ServiceUnavailable) -> null
     ex.hasStatusCode(411) -> null
     ex.hasStatusCode(411, 403) -> null
     ex.hasNotStatusCode(423) -> null
     ex.hasNotStatusCode(503, 422) -> null
     else -> null
   }
 }
```

### Message
Since messages are an important part of bots, here are some features to manage some event about a message
For you can easily handle the following events on your message
- ReactionAddEvent
- ReactionRemoveEvent
- ReactionRemoveAllEvent
- MessageDeleteEvent
- MessageUpdateEvent
- MessageBulkDeleteEvent

For this here is how to do it : 
```kotlin
message.events {
  update { 
     it.channel.createMessage("The message is updated")
  }
  delete {
     it.channel.createMessage("The message is deleted")
  }
  deleteOnly {
      it.channel.createMessage("The message is deleted by a selection")
  }
  deleteBulk {
      it.channel.createMessage("The message is deleted with other messages")
  }
  reaction {
    if(it is ReactionAddEvent){
           it.channel.createMessage("The message has a new reaction")
     }
     else if(it is ReactionRemoveEvent){
            it.channel.createMessage("The message has lost a reaction")
     }
  }
  reaction(cross) {
      if(it is ReactionAddEvent){
          it.channel.createMessage("The message has a new cross reaction")
      }
      else if(it is ReactionRemoveEvent){
           it.channel.createMessage("The message has  lost a cross reaction")
      }
  }
  reactionAdd {
      it.channel.createMessage("The message has a new reaction")
  }
  reactionAdd(cross) {
       it.channel.createMessage("The message has a new cross reaction")
  }
  reactionRemove {
       it.channel.createMessage("The message has lost a reaction")
  }
  reactionRemove(cross) {
       it.channel.createMessage("The message has lost a cross reaction")
  }
  reactionRemoveAll {
       it.channel.createMessage("The message has lost all reactions")
  }
  stop { 
       it?.channel?.createMessage("The message has lost his listener for events")
  }
}
```


kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/checks/MemberChecks.kt
```diff
@@ -28,7 +28,7 @@ public fun hasPermission(perm: Permission): suspend (Event) -> Boolean {
             return false
         }
 
-        return if (member.asMember().hasPermission(perm)) {
+        return if (member.asMember().hasPermissions(perm)) {
```
gdude2002: If you're going to change all calls to `hasPermission`, which seems inadvisable for performance reasons, surely you'd deprecate it? I'm not sure why it was changed here, as the checks only check a single permission.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/message/MessageEventManager.kt
```diff
@@ -0,0 +1,267 @@
...
+public open class MessageEventManager(
+    public val message: MessageBehavior,
+    private val timeout: Long? = DEFAULT_TIME_LISTENING
+) {
+
+    public companion object {
+        /**
+         * Default time before stop the listening of events.
+         */
+        public const val DEFAULT_TIME_LISTENING: Long = 1000L * 60L * 5L
```
gdude2002: We generally put constants like this outside the class, you may have seen this elsewhere...

```kotlin
// 5 minutes, default time to wait for message events before stopping
private const val DEFAULT_TIME_LISTENING: Long = 1000L * 60L * 5L

// ...

public open class MessageEventManager(
```

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/message/MessageEventManager.kt
```diff
@@ -0,0 +1,267 @@
...
+ */
+public open class MessageEventManager(
+    public val message: MessageBehavior,
+    private val timeout: Long? = DEFAULT_TIME_LISTENING
+) {
+
+    public companion object {
+        /**
+         * Default time before stop the listening of events.
+         */
```
gdude2002: We'll need to review the documentation comments separately as they don't read very well — this isn't your fault, you're not a native English speaker, but we do have an expected style across the project.

I'll probably redo them myself at some point.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/message/MessageEventManager.kt
```diff
@@ -0,0 +1,267 @@
...
+    /**
+     * Job running to listen to events.
+     */
+    public var job: Job? = null
+        private set
+
+    /**
+     * `true` if the manager listening the events, `false` otherwise.
+     */
+    @Volatile
```
gdude2002, line 45: Isn't this just a signifier for Java users? We don't support Java users, we're a Kotlin-only library, and eventually I expect we'll end up multiplatform so I'm not really sure annotations like this are necessary.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/message/MessageEventManager.kt
```diff
@@ -0,0 +1,267 @@
...
+        event {
+            val deleteEvent = when (it) {
+                is MessageDeleteEvent -> it
+                is MessageBulkDeleteEvent -> {
+                    val msg = if (message is Message) message else null
+                    MessageDeleteEvent(message.id, it.channelId, it.guildId, msg, it.kord, it.shard, it.supplier)
+                }
+                else -> return@event
+            }
+            block(deleteEvent)
```
gdude2002, line 150: This block of code is too dense, but this isn't the only one — take a look over all the code you changed and add more blank lines to it! Dense code is harder to read and harder to review.

In this case, I'd expect a blank line before the `else` branch in this `when` block, and before the `block` call above.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/message/MessageEventManager.kt
```diff
@@ -0,0 +1,267 @@
...
+     * Start the listening if necessary.
+     * @return `true` if the listening start, `false` if he is already started
+     */
+    public open fun start(): Boolean {
+        if (listening || job != null) return false
+
+        job = kord.launch {
+            val condition = getCondition()
+            listening = true
+            var isDelete = false
```
gdude2002, line 213: This entire block is too dense, but you also probably want to re-order the above lines so that they make more sense. I'd do something like this:

```kotlin
val condition = getCondition()
var isDelete = false

listening = true
```

gdude2002: Below this, I'd expect a blank line between the `var` and the first line of the conditional, a blank line after the `while` block, and a grouping of the assignments if possible (or at least separating them from the function call)

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/message/MessageEventManager.kt
```diff
@@ -0,0 +1,267 @@
...
+        when (it) {
+            is ReactionAddEvent -> id == it.messageId && it.userId != kord.selfId
+            is ReactionRemoveEvent -> id == it.messageId && it.userId != kord.selfId
+            is ReactionRemoveAllEvent -> id == it.messageId
+            is MessageDeleteEvent -> id == it.messageId
+            is MessageUpdateEvent -> id == it.messageId
+            is MessageBulkDeleteEvent -> id in it.messageIds
+            else -> false
+        }
+    }
```
gdude2002, line 275: In this block, I'd expect a blank line after the `var` and before the `else` branch of this `while` block. You see what I'm getting at with these density comments?

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Channel.kt
```diff
@@ -1,11 +1,18 @@
+@file:JvmMultifileClass
+@file:JvmName("ChannelKt")
```
gdude2002: I can't think of a reason to add these annotations, and they don't really have much to do with the PR. We do not support Java users, we're a Kotlin-only library.

This comment applies to the other files with this type of annotation as well.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Channel.kt
```diff
@@ -1,11 +1,18 @@
...
 import dev.kord.core.entity.Webhook
 import dev.kord.core.entity.channel.GuildMessageChannel
 import dev.kord.core.firstOrNull
 import dev.kord.rest.Image
-import mu.KotlinLogging
+
+/**
+ * Logger of the class.
+ */
+private val LOG = classLogger()
```
gdude2002: This is literally the same as the previous `logger` vals I had sprinkled around. Factoring that out into a function is not very useful (especially since the loggers will be created for the wrong scope), and those loggers were intentionally created within functions because they were specific to those functions (ideally with a matching name, but I didn't get around to doing that for the utils functions yet).

This applies to all instances where you've changed how the loggers work. These changes are out of scope for this PR, and the loggers were placed that way by intentional design.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Enum.kt
```diff
@@ -0,0 +1,18 @@
...
+ * Retrieve an enum by his name.
+ * @receiver Enum in which the search will process.
+ * @param name Name of enum searched.
+ * @return The enum value if the name is found, null otherwise.
+ */
+public fun <T : Enum<*>> KClass<T>.findByName(name: String, ignoreCase: Boolean = true): T? {
+    return if (name.isEmpty()) {
+        null
+    } else java.enumConstants.find { it.name.equals(name, ignoreCase) }
+}
```
gdude2002: I still don't really see the point of this one. Additionally, it relies too much on Java's internals to work, instead of using, for example, [Kotlin's enumValues function](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/enum-values.html).

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Koin.kt
```diff
@@ -11,8 +14,6 @@ public fun Koin.module(
     moduleDeclaration: ModuleDeclaration
 ): Module {
     val module = org.koin.dsl.module(createdAtStart, override, moduleDeclaration)
-
     loadModules(listOf(module))
-
```
gdude2002: Don't remove spacing that someone else put in, it makes the code objectively harder to read.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Member.kt
```diff
@@ -0,0 +1,90 @@
...
+ * @param perms The permissions to check for
+ * @return `true` if the collection is empty,
+ * or the [Member] has the permissions (also if he has the administrator permission),
+ * `false` otherwise
+ */
+public suspend fun Member.hasPermissions(perms: Collection<Permission>): Boolean {
+    if (perms.isEmpty()) return true
+    val permissions = getPermissions()
+    return perms.all { it in permissions }
+}
```
gdude2002, line 85: Again, this function is too dense. It should be 5 lines, not 3.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Member.kt
```diff
@@ -0,0 +1,90 @@
...
+ * always return `true` if the member has the `Administrator` permission in one of their roles.
+ *
+ * @receiver The [Member] check permissions for for
+ * @return Whether the [Member] has the given permission, or the Administrator permission
+ */
+public suspend fun Member.hasPermission(perm: Permission): Boolean =
+    perm in getPermissions()
+
+/**
+ * Convenience function to check whether a guild member has the permissions.
```
gdude2002, line 54: You wanted `the given permissions` here, I think. `the permissions` doesn't read well in English.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Member.kt
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Convenience function to check whether a guild member has a permission.
+ *
+ * This function only checks for permissions based on roles, and does not deal with channel overrides. It will
+ * always return `true` if the member has the `Administrator` permission in one of their roles.
+ *
+ * @receiver The [Member] check permissions for for
+ * @return Whether the [Member] has the given permission, or the Administrator permission
+ */
+public suspend fun Member.hasPermission(perm: Permission): Boolean =
```
gdude2002: See my previous comment from the permissions-related checks.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Message.kt
```diff
@@ -1,27 +1,164 @@
...
+
+/**
+ * Logger of the class.
+ */
+private val LOG = classLogger()
 
+/**
+ * Time to delete information.
+ */
 private const val DELETE_DELAY = 1000L * 30L  // 30 seconds
```
gdude2002, line 24: Private values don't need to be documented, as they won't show up in the generated documentation. It's fine to document them if they're confusing or otherwise not obvious, but I'd say it's unnecessary in most of these cases.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Message.kt
```diff
@@ -1,27 +1,164 @@
...
+
+/**
+ * Deletes a message after a delay.
+ *
+ * This function **does not block**.
+ *
+ * @param millis The delay before deleting the message, in milliseconds.
+ * @return Job spawned by the CoroutineScope.
+ */
+public fun MessageBehavior.delete(millis: Long, retry: Boolean = true): Job {
```
gdude2002, line 60: You renamed this, but didn't leave behind a deprecated function (I'd just set it to a warning for now) with the old name. Make sure you do that to avoid breaking people's code too frequently.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Message.kt
```diff
@@ -1,27 +1,164 @@
...
+
 /** ID of the message author. **/
-public val MessageData.authorId: Snowflake get() = author.id
+public val MessageData.authorId: Snowflake
+    get() = author.id
 
 /** Is the message author a bot. **/
-public val MessageData.authorIsBot: Boolean get() = author.bot.discordBoolean
+public val MessageData.authorIsBot: Boolean
+    get() = author.bot.discordBoolean
```
gdude2002, line 156: These functions fit on one line, so there was no real need to drop them to a new line.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Message.kt
```diff
@@ -97,9 +232,7 @@ public suspend fun Message.respond(
 
     return if (useReply) {
         reply { innerBuilder() }
-    } else {
-        channel.createMessage { innerBuilder() }
-    }
+    } else channel.createMessage { innerBuilder() }
```
gdude2002: No, the other branch uses braces so this branch should as well.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Message.kt
```diff
@@ -183,27 +269,23 @@ public suspend fun Message.requireChannel(
     deleteOriginal: Boolean = true,
     deleteResponse: Boolean = true
 ): Boolean {
-    val topRole = if (getGuildOrNull() != null) {
-        this.getAuthorAsMember()!!.getTopRole()
-    } else {
+    val topRole = if (getGuildOrNull() == null) {
         null
-    }
+    } else getAuthorAsMember()!!.getTopRole()
```
gdude2002: Again, braces please. Also, I'm not sure why we're getting rid of `this` everywhere — I find it often makes things easier to understand, but I don't seem to be able to lint for it.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Message.kt
```diff
@@ -219,18 +301,19 @@ public suspend fun Message.requireChannel(
  * @return true if the message was posted in an appropriate context, false otherwise
  */
 public suspend fun Message.requireGuildChannel(role: Role? = null): Boolean {
-    val topRole = if (author != null && getGuildOrNull() != null) author!!.asMemberOrNull(getGuild().id) else null
+    val author = this.author
+    val guild = getGuildOrNull()
+    val topRole = if (author != null && guild != null) {
+        author.asMemberOrNull(guild.id)
+    } else null
```
gdude2002: Gotta get them braces.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Message.kt
```diff
@@ -249,17 +332,19 @@ public suspend fun Message.requireGuildChannel(role: Role? = null): Boolean {
  * @return true if the message was posted in an appropriate context, false otherwise
  */
 public suspend fun Message.requireGuildChannel(role: Role? = null, guild: Guild? = null): Boolean {
-    val topRole = if (author != null) guild?.getMember(this.author!!.id)?.getTopRole() else null
+    val author = this.author
+    val topRole = if (author != null) {
+        guild?.getMember(author.id)?.getTopRole()
+    } else {
+        null
+    }
```
gdude2002, line 342: See? Braces! Perfect!

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Misc.kt
```diff
@@ -0,0 +1,30 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import mu.KLogger
+import mu.KotlinLogging
+
+/**
+ * Neutral and default logger.
+ */
+public fun classLogger(): KLogger = KotlinLogging.logger() {}
```
gdude2002: I made a comment about this earlier, but I really don't see the point of doing this. It doesn't add any utility, and currently your loggers aren't being instantiated in the right scope. It's better to just do this as explained in my other comment.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_RestRequest.kt
```diff
@@ -0,0 +1,49 @@
...
+    return codes.any { it == code }
+}
+
+/**
+ * Check if the [RestRequestException] has not the same HTTP code of a [HttpStatusCode].
+ * @receiver Exception from a request with rest interface.
+ * @param status Status to compare the code of the exception.
+ * @return `true` if there is none [HttpStatusCode] with the same HTTP code, `false` otherwise.
+ */
+public fun RestRequestException.hasNotStatus(vararg status: HttpStatusCode): Boolean =
```
gdude2002: Surely we can just do `!hasStatus()` or `hasStatus().not()` instead of writing a separate function for it.

It's fine to keep, I just don't entirely see the point.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Scheduler.kt
```diff
@@ -1,17 +1,22 @@
+@file:JvmMultifileClass
```
gdude2002: Most of your changes in this file involve these annotations, making the file harder to read, or replacing logger creations. I don't see the need or point of those changes, and they're out of scope of the PR.

I support the change to `forEach` for the jobs, though.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_String.kt
```diff
@@ -0,0 +1,75 @@
+@file:JvmMultifileClass
```
gdude2002: Almost none of the changes to this file were necessary, I'd say. The boolean parsing was fine and, as I mentioned on Discord, is going to be replaced when we eventually get around to i18n support.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Time.kt
```diff
@@ -34,8 +38,7 @@ public fun Duration<IsoUnit>.toSeconds(): Long {
  */
 @Suppress("MagicNumber")  // These are all time units!
 public fun java.time.Duration.toHuman(): String? {
-    val parts = mutableListOf<String>()
```
gdude2002: While I appreciate the refactoring, your version of this function doesn't give the same output.

gdude2002 (changes requested):
Not bad, but quite a few things to look into. I didn't leave a comment for every documentation comment because almost all of them need to be rewritten — that isn't your fault, you're not a native English speaker, so I'll handle that when I can (assuming I have access to push to your branch).

I do feel like you're prioritizing Java users with some of your changes, which is entirely out of scope of our project. Java users have plenty of Discord libraries already, many of which are better suited to Java than ours, which already assumes Kotlin in many, many situations. Annotations like `@Volatile`, `@JvmStatic`, etc therefore probably aren't necessary.

Additionally, you don't seem to be following the style conventions of the project. While this is partially my fault (I haven't had a chance to mess with the linter yet), I'd appreciate if you spaced your code out more and didn't "correct" existing code that's already there to match your own idea of how code should be styled.

Overall, though, pretty good, lots of useful things in here. I appreciate the tests as well. Once you've addressed these comments, we can take a look at the doc comments.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/checks/MemberChecks.kt
Distractic: Indeed, at first the hasPermission method was deprecated, but in the end I thought it was a good idea to leave it in addition to the methods to check multiple permissions. Therefore I will correct to put the version with a simple argument

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/message/MessageEventManager.kt
Distractic: I put this variable in the class since it is specific to this class
But if this is the convention for the project, I respect it and move the variable

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/message/MessageEventManager.kt
Distractic: The "Volatile" annotation is not for java users, in fact it is an annotation taken into account by Kotlin for JS, JVM, etc. (cf. https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-volatile/). 
I use it because the variable can be both modified or watched by 2 different threads. 
After that it's just to be sure that the information is correctly obtained and updated, it can be a bit overkill

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Enum.kt
Distractic: Yes quite. It was just to add a pre-made function, but you're right, The class is removed

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Message.kt
Distractic: So maybe you have a big screen, but trust me that for small screens (like mine) having it on the next row allows you to instantly see which method is called

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Message.kt
Distractic: I find your nemesis

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/message/MessageEventManager.kt
gdude2002: This comment hasn't yet been addressed

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Message.kt
gdude2002: My margin is set to 120 chars — are you using an 80-char TTY? As [Torvalds himself says](https://www.theregister.com/2020/06/01/linux_5_7/):

> People with restrictive hardware shouldn't make it more inconvenient for people who have better resources. Yes, we'll accommodate things to within reasonable limits. But no, 80-column terminals in 2020 isn't "reasonable" any more as far as I'm concerned. People commonly used 132-column terminals even back in the '80s, for chrissake, don't try to make 80 columns some immovable standard.

gdude2002:
Aside from the handful of comments I just made, I think this PR is pretty good. We'll need to do a documentation pass later, but that's not super important right now.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Message.kt
Distractic: Indeed, I had not seen that Intellij put the margin in the editor
So, it's corrected

gdude2002: You may have forgotten to push that one. But I'm starting to think a lot of these stylistic changes are better handled with a more specific linter configuration, so you can leave it for now.

gdude2002:
I think overall this is looking pretty good now. There are still a few things left to do — specifically, the documentation comments need to be rewritten, and the actual documentation site needs updating, but I'm happy to do both of those myself.

While there are stylistic changes that I don't agree with, I think that's a problem better solved by using a linter. With this in mind, I'm going to merge your pull request and redo the documentation myself.

Good work!

PR #18, opened 2021-01-13
Integer and Long converters

bosukas:


kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/converters/impl/NumberConverter.kt
```diff
@@ -39,4 +39,4 @@ public class NumberConverter(
 
     override suspend fun toSlashOption(arg: Argument<*>): OptionsBuilder =
         IntChoiceBuilder(arg.displayName, arg.description).apply { required = true }
-}
+}
```
gdude2002: You removed the newline at the end of this file for some reason — don't do that, it'll fail the build. All files must have a newline at the end.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/converters/Functions.kt
```diff
@@ -766,6 +862,36 @@ public fun Arguments.numberList(
...
+    arg(displayName, description, LongConverter(radix).toMulti(required, signatureTypeString = "numbers"))
+
+/**
+ * Create an integer converter, for lists of arguments.
+ *
+ * @param required Whether command parsing should fail if no arguments could be converted.
+ *
+ * @see IntConverter
+ */
+public fun Arguments.intList(
```
gdude2002: The functions in each region are alphabetically sorted, and yours should be no different. Please ensure that, within each region, the functions are sorted alphabetically — that means it'll be far, far easier to find what one is looking for.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/converters/Functions.kt
```diff
@@ -286,6 +309,13 @@ public fun Arguments.optionalMessage(
  *
  * @see NumberConverter
  */
+@Deprecated(
```
gdude2002, line 339: Please add `DeprecationLevel.WARNING` to each deprecation, to be explicit.

gdude2002 (changes requested):
Pretty good, just a couple small changes.

Please use the linter, by the way — it's the `detekt` gradle task.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/converters/impl/NumberConverter.kt
bosukas: hmm, not sure why this happened, must of been an accident, ill fix this

PR #21, opened 2021-01-19
fix: Remove listener when channel or guild is deleted

Distractic:
# Why

In the previous version of the event system for messages. Some event that deleted the message was not supported _(ChannelDeleteEvent & GuildDeleteEvent)_.

This version fixes that and allows not to keep the event listener when the channel or guild is deleted


gdude2002:
I think instead of maintaining this, we should prefer making use of Kord's live entities. While our abstraction does provide a few small conveniences, I think we'd be better off letting Kord maintain a feature like this.

PR #24, opened 2021-02-27
Global & local command checks

bosukas:


bosukas:
Adding the documentation for the variables tomorrow so that it can build

gdude2002:
I'm overall happy with this PR and its quality - the only thing that needs doing is the documentation. I'll handle that. Thanks!

PR #38, opened 2021-05-14
Cooldowns

bosukas:
#36


gdude2002:
I'm having a hard time really understanding how or why this works, but I honestly just can't get my head around this kind of code.

Either way, looks like you've got some conflicts to resolve there

bosukas:
The idea is that there are different cooldown types, like guild, user and channel cooldowns, as well as other custom cooldowns users may want like different cooldowns according to a bot's own levelling system (e.g. level 1 = 15 second cooldown, level 2 = 10 second cooldown, so on) so I've created a `CooldownType` class for this. 

Each cooldown type should generate a unique key that represents its cooldown type and who its for. For example, in the `UserCooldown`, the unique key is `U:${user_id}`, and in the `GuildCooldown`, the unique key is `G:${guild_id}`. The keys are also nullable if the unique key could not be generated. For example, the `GuildCooldown` will return null if the message was not from a guild.

Since I cant possibly code out every single possible cooldown a user may want to have, users can create their own cooldown types and I have allowed the user to define what cooldowns types are checked and applied by using the `priority` function in the `CooldownsBuilder`. This is just a list of cooldown types that the user wants to use along with what cooldowns should be checked first. This list of cooldown types is then iterated over when checking for cooldowns

```kt
for (cooldownType in extension.bot.settings.cooldownsBuilder.priority.invoke()) {
    // the unique key for the cooldown type + who its for. if its null, move onto the next cooldown type we should check
	val key = cooldownType.getCooldownKey(event) ?: continue

    // get the time left for the cooldown (if any)
	val timeLeft = cooldowns.getCooldown(key)
    
    // get the duration that this cooldown is meant to last for
	val cooldownDuration = cooldownBody.invoke(cooldownType)

    when {
    	cooldownDuration == null -> continue // this cooldown type does not have a duration
        timeLeft == null -> cooldowns.setCooldown(key, cooldownDuration)
        else -> if (timeLeft < cooldownDuration) {
        	throw CommandException("You must wait another ${timeLeft.inSeconds} seconds")
        } else {
            cooldowns.setCooldown(key, cooldownDuration)
   		}
	}
}
```

Users can define what the actual cooldown is for different cooldown types by using the cooldown dsl in the command. For example, 
```kt
cooldowns {
	when(it) {
            is UserCooldown -> 5.seconds
            else -> null
	}
}
```

Each command has a cooldown object, which keeps track of the cooldowns. Users can use their own implementation of the cooldown object if they want to for whatever reason, by using the `implementation` function on the `CooldownsBuilder`.

The only problem for this is that while technically multiple cooldowns are supported for a single command, it gets a little weird with how it works but I'm not sure how to overcome that.

Another problem is that if the user has not registered a certain cooldown type on the priority and try to use in a command, it will not work and just ignore it. 

I've tried to make this as customizable as possible.

Does that help a bit?

bosukas:
Another thing I'm thinking of adding is the ability to import cooldowns on bot start-up and export cooldowns before the bot shuts-down, however I'm not too sure how this would work right now

bosukas:
Syntax for the cooldown body has changed slightly. I have decided that commands should not have multiple cooldowns.
Cooldowns are now defined like this (from the test command in the test extension):
```kt
command {
            name = "cooldown-test"
            description = "Tests the User cooldown."

            cooldowns<UserCooldown> {
                Duration.seconds(30)
            }

            action {
                message.respond("There is no cooldown!")
            }
        }
```

gdude2002:
This is a fairly old PR at this point - ultimately, I don't think either of us have the skills necessary to get this feature done and reviewed. I massively appreciate the work you've put in, though!

I'll be closing this for now, to clean things up.

PR #44, opened 2021-05-28
Replace jump url method for variable instead

bosukas:
Just a small change that replaces the `getUrl` method for a `jumpUrl` variable instead, I feel like this would be better as it doesn't require a suspending function.

### DIFF
--- a/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Message.kt
+++ b/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Message.kt
@@ -1,6 +1,7 @@
 package com.kotlindiscord.kord.extensions.utils
 
 import com.kotlindiscord.kord.extensions.commands.CommandContext
+import dev.kord.common.entity.DiscordPartialMessage
 import dev.kord.common.entity.MessageFlag
 import dev.kord.common.entity.Snowflake
 import dev.kord.core.behavior.MessageBehavior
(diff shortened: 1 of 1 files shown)


gdude2002:
I don't think this is an appropriate use of a `val` with getter, as it's not clear that work's being done. I'd accept a solution that uses `by lazy {}`, provided you can explain why it's a better approach than a function.

If your approach to getting this to work in a `val` is possible, then it can just as well be put into a non-suspending function either way.

bosukas:
The main thing I wanted is to make it non-suspending, I just thought that I could also fit it into a variable since it only does string concatenation. A non-suspending function would also work fine, I'll change this now

gdude2002:
At this point, why rename it? Might as well just keep the old name and avoid the need for a deprecation.

bosukas:
Personally, I feel as though `jumpUrl` is better than just `url`, as it will help people find and recognise this method easier through ide auto-complete because there's another keyword `jump`. It also gives more information about what this url is (that its a jump url for the message), rather than just `url`.

It's a pretty small change but I feel like its useful. However if you still want me to change it back I can.

gdude2002:
It seems like a fairly minor reason for a breaking change. Where is it usually called a jump URL? I've never seen that, at least not that I can remember.

bosukas:
I guess its not officially called a jump url, but (from what i've seen) its often referred to as a jump url, where when you click on the link you 'jump' to the message. I've just always known it as the jump url, i've never heard it just be the 'url' for a message

bosukas:
It might also be because the first discord wrapper I ever used was JDA, and they called it a jump url

gdude2002:
hmm, I guess that's OK then, if it's something people are familiar with.

gdude2002 (approved):

PR #64, opened 2021-07-21
insert semicolon in Lclass;method for mixin target

spaceclouds42:
Fixes an issue where Mixin targets for methods would not have the `;` between class and method descriptors


gdude2002 (approved):

gdude2002:
Yaaaaaaaay, thanks again haha

PR #95, opened 2021-09-29
Add self member tools

DRSchlaubi:
- Add Guild.selfMember()
- Add GuildBehavior.botHasPermissions()
- Add GuildChannel.botHasPermissions()﻿

### DIFF
--- a/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Guilds.kt
+++ b/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Guilds.kt
@@ -39,6 +39,5 @@ private suspend fun GuildBehavior.botHasPermissions(channel: GuildChannel?, requ
             // slow lane: if guild was cached we fetch it to retrieve permissions field
             ?: withStrategy(EntitySupplyStrategy.rest).asGuild().permissions!!
 
-
     return requiredPermissions in effectivePermissions
 }
--- a/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Guilds.kt
+++ b/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Guilds.kt
@@ -1,0 +1,44 @@
+package com.kotlindiscord.kord.extensions.utils
+
+import dev.kord.common.entity.Permission
+import dev.kord.common.entity.Permissions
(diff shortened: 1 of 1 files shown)


kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Guilds.kt
```diff
@@ -0,0 +1,43 @@
...
+public suspend fun GuildBehavior.botHasPermissions(vararg requiredPermissions: Permission): Boolean =
+    botHasPermissions(null, Permissions(requiredPermissions.asIterable()))
+
+private suspend fun GuildBehavior.botHasPermissions(channel: GuildChannel?, requiredPermissions: Permissions): Boolean {
+    val effectivePermissions =
+        channel?.run { permissionsForMember(selfMember()) }
+        // fast lane: this guild was fetched through rest
+            ?: asGuild().permissions
+            // slow lane: if guild was cached we fetch it to retrieve permissions field
+            ?: withStrategy(EntitySupplyStrategy.rest).asGuild().permissions!!
```
gdude2002, line 40: We prefer that API method wrappers provided by KordEx adhere to the caching settings provided by the `ExtensibleBotBuilder`, which defaults to the caching REST supplier - wouldn't `asGuild()` already hit cache?

DRSchlaubi: > wouldn't asGuild() already hit cache?

Yes that's the issue because if we take a look at Guild.permissions
https://github.com/kordlib/kord/blob/28fab1af5f768d666e763aebe5a6910629602c29/core/src/main/kotlin/entity/Guild.kt#L96-L102

guilds are normally cache upon gateway connect, so the cached version won't have the permissions field, which is why we fetch permissions through rest if it doesn't

gdude2002: Wouldn't checking permissions this way be inherently faulty anyway, though? I mean, this isn't going to check what perms the bot's roles have, right?

DRSchlaubi: yep you're right

DRSchlaubi:
I totally didn't close this to make the new PR count for Hacktoberfest

PR #96, opened 2021-09-29
Cleanup gradle files

DRSchlaubi:
The Gradle configuration showed a lot of duplicated code, this PR aims to fix that﻿


gdude2002:
This looks considerably better, but can you quickly ELI5 what you've done here? It looks like you've created three gradle plugins that handle the common tasks, right?

DRSchlaubi:
I have no idea what ELI5 means so I assume you want a quick TL;DR

If you craete a gradle.kts in buildSrc Gradle will generate a plugin implimentation for you, which applies that buildscript as a plugin.

So here q quick explanation about what each one does and how you can configure it

- `kordex-module`:
  desc: top level plugin
  configures: kotlin (jvmtarget and java plugin target), detekt config, sources and javadoc jars, explicitApi mode
  config: if you don't target java 8 or want to disable explicitApi, you have the Kordex extension

```kotlin
kordex {
    jvmTarget.set("9")
    javaVersion.set(JavaVersion.VERSION_1_9)
    useExplicitApiMode.set(false)
}
```

- `tested-module`
  configures: test task (**THIS DOESN'T ADD JUNIT DEPS**, because you can't use dependency catalogs in buildSrc)

- `ksp-module`
 configures: ksp source sets (this adds the ksp generated output to the main source set, ~~it doesn't add the plugin for ksp~~ though)
- `dokka-module`
  configures: dokkaHtml task
  configuration: If you want to set a moduleName, or include md files

```kotlin
dokkaModule {
    moduleName.set("Kord Extensions: Annotation Processor")
    includes.add("packages.md")
}
```

- `published-module.`
  desc: configures publishing
  configures: signing and maven publish tasks

gdude2002:
Hmm, okay, that makes a lot of sense. Any reason the KSP plugin can't be applied by `ksp-module`? There'll be no KSP output without the plugin regardless.

DRSchlaubi:
> Hmm, okay, that makes a lot of sense. Any reason the KSP plugin can't be applied by `ksp-module`? There'll be no KSP output without the plugin regardless.

no reason it can't, it just wasn't applied to all modules having the sourceSets

gdude2002:
You can chalk that up to poor copypasta, lmao

DRSchlaubi:
> You can chalk that up to poor copypasta, lmao

Okay then'll remove the dokka-module from all modules not having the plugin applied and add the plugin directly in the module

gdude2002:
Good idea, appreciate that

DRSchlaubi:
Okay one last problem seems to be that intellij doesn't pick up the kotlin config correctly I will try to fix that real quick

gdude2002:
We good?

DRSchlaubi:
We should be

gdude2002:
🚀 

Thanks for looking into this!

Issue #86, opened 2021-09-23
Flexible state handling for slash commands and components


ByteAlex:
# Summary

In the current implementation of KordEx, the entire state of slash commands and components is only accessible in the virtual environment. Distributed setups with multiple bot instances are facing an issue there, commands/components which are registered/created on one VM can not properly be accessed/used on other VMs.
This is a major and blocking issue for my use case.

# Current Behaviour
The state "internal ids" are stored in a local hashmap, which is not VM agnostic and losing all information with shutdown of the VM.

# Suggestions / Preferred Behaviour
I can imagine two ways of handling this:
a) Give commands & components an own discord-agnostic id, which is constant across VMs and throughout restarts. This would eventually make it possible to use KordEx in distributed environments.
b) Install an interface to store state, which can default to a "local" (HashMap or similar) implementation, but it pluggable to switch to a networke'd state-store like redis for example (to stay with simple key-value bindings). Though I also can imagine a SQL table for slash commands, since those should be more persistant.

Possibly a mix of both would result in the best dev-experience and have a solid feature set for distributed bots.

gdude2002:
It's certainly an interesting idea.

> Give commands & components an own discord-agnostic id, which is constant across VMs and throughout restarts. This would eventually make it possible to use KordEx in distributed environments.

Technically speaking, this is already done. Components have unique IDs _by default_ (they're given a stringified UUID), and commands must have unique names in their own scopes (guild/global and message/slash/user) - which effectively is an ID of sorts. The Snowflake to command mapping only exist because Discord keys them by Snowflake. 

I'm not sure what you specifically want from such a system, given the above.

> Install an interface to store state, which can default to a "local" (HashMap or similar) implementation, but it pluggable to switch to a networke'd state-store like redis for example (to stay with simple key-value bindings). Though I also can imagine a SQL table for slash commands, since those should be more persistant.

I see no problem with doing this, but I get the feeling that the current access methods might not suit your use-case. Can you be more specific with what you need?

---

My assumption from your questions on Discord was that what you actually wanted was a way to register reusable command/component actions/builders to IDs, which would mean you'd have your own registry of actions/builders, and you'd just tell the commands DSL which IDs to use. That seems like a more logical approach to me.

ByteAlex:
Thank you very much for your quick answer!

[...]

PR #97, opened 2021-09-30
First approach for a generic registry storage interface

ByteAlex:
First approach to solve the issue in #86 with a generic storage interface, - which probably, in its current state, will not cover the entire use case yet.


kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
```diff
@@ -225,9 +229,9 @@ public open class ApplicationCommandRegistry : KoinComponent {
 
                 if (existingCommand != null) {
                     when (commandObj) {
-                        is MessageCommand<*> -> messageCommands[existingCommand.id] = commandObj
-                        is SlashCommand<*, *> -> slashCommands[existingCommand.id] = commandObj
-                        is UserCommand<*> -> userCommands[existingCommand.id] = commandObj
+                        is MessageCommand<*> -> messageCommands.upsert(existingCommand.id, commandObj)
```
gdude2002: You can implement this operator, if you didn't know. I think I'd recommend it haha

gdude2002:
Looks pretty good so far - the build is failing, obviously, with compile errors, so that's probably something you need to address next.

I think you should implement the get/set operator functions that map-like access uses, since, well, this is functionally pretty similar - but if you can't implement them in a suspending way, don't worry about it.

I think the function naming is very odd. `read`, `upsert` and `delete` aren't natural for collection usage, and I'd prefer `get`, `put`/`set` and `remove`. But, again, if those are reserved and you can't make them suspending, don't worry about it.

ByteAlex:
Suspend operator won't work for:
* set
* get

Remove doesn't have an operator fun, afaik.

I changed the naming to `set`, `get`, `remove` though.

My idea behind the naming was taking the `Registry` as a `Repository`, which supports `CRUD`, but since we do not have a specific create/update, I named it `upsert`.

Other than that, the compile errors are only suspend funs called in forEach loops (if I saw correctly), which needs some minor refactoring. This can be done at any time.

But before going deeper into code & verification, I want to outline the idea more specifically.
One thing I noticed, is that the `Registry` needs a way to put values in the local cache, without the "unknown" snowflake/uuid.

I.e. if you locally register/create the commands, they should register themselves with their unique `name` in the registry, and if the registry doesn't know the IDs for those names, it would have to reregister them at discord and then upsert the ID/Command entries to the registry.

Do you agree with this approach?

gdude2002:
I'm not entirely sure if I follow you there - commands registered at startup time need to be "wired" into their respective discord-side IDs for handling, but Discord does also key based on unique names too. For that reason I'm not actually sure what you're trying to solve by changing the current "register the commands with discord at startup as is needed" approach?

ByteAlex:
If we assume the registry has state across multiple processes, it already knows some "command-id -> name" correlations.
With the startup of a new process, we need to find out which correlations it does not know, to register them at discord.
Also in my AbstractDeconstructingApplicationCommandRegistryStorage commands are reconstructed from a name, so each command needs to register its name -> ApplicationCommand object at startup, so the registry can reconstruct those objects from id/name.

gdude2002:
I don't think that actually matters - the current system uses the names while registering them in the first place. All you need is to support name-based lookups in your registry implementation to see if they're already there and, presumably, update them on Discord if they are there but have changes, right?

Otherwise, what you're doing is dispatching dynamically-created command objects based on the IDs presented in teh events.

ByteAlex:
I pushed some changes, which go quite deep into Kordex' architecture. Could you please review those before I continue to fight this battle?

gdude2002:
Seems okay so far, I'd say.

ByteAlex:
I think what I have now would probably work, I just need to think of a smart way of syncing commands on init `if needed`.

ByteAlex:
Testing the default implementation:

Expected output:

```log
20:31:06.928 [DefaultDispatcher-worker-1] INFO com.kotlindiscord.kord.extensions.commands.application.ApplicationCommandRegistry - Global application commands: 9 to add / 0 to update / 0 to remove
20:31:08.056 [ktor-cio-dispatcher-worker-3] INFO com.kotlindiscord.kord.extensions.commands.application.ApplicationCommandRegistry - Finished synchronising global application commands
```

Result:

```log
20:46:09.110 [DefaultDispatcher-worker-1] INFO com.kotlindiscord.kord.extensions.commands.application.ApplicationCommandRegistry - Global application commands: 9 to add / 0 to update / 0 to remove
20:46:09.962 [ktor-cio-dispatcher-worker-1] INFO com.kotlindiscord.kord.extensions.commands.application.ApplicationCommandRegistry - Finished synchronising global application commands
```

Expected commands were synced and working.

I tried a simple "FileRegistry" for the `StorageAwareApplicationCommandRegistry` and it worked aswell!

```kotlin
import com.kotlindiscord.kord.extensions.commands.application.ApplicationCommand
import com.kotlindiscord.kord.extensions.registry.AbstractDeconstructingApplicationCommandRegistryStorage
import com.kotlindiscord.kord.extensions.registry.RegistryStorage
import com.kotlindiscord.kord.extensions.utils.runSuspended
import kotlinx.coroutines.Dispatchers
import kotlinx.coroutines.flow.Flow
import kotlinx.coroutines.flow.asFlow
import java.io.File

class FileRegistryStorage<T : ApplicationCommand<*>>(private val file: File) :
    AbstractDeconstructingApplicationCommandRegistryStorage<T>() {

    init {
        file.createNewFile()
    }

    private var fileData: MutableMap<String, String> = readFile()

    private fun readFile(): MutableMap<String, String> {
        fileData = mutableMapOf()
        file.readLines()
            .asSequence()
            .filter { it.isNotEmpty() }
            .filter { it.contains("=") }
            .forEach {
                val data = it.split(Regex("="), limit = 2)
                fileData[data[0]] = data[1]
            }
        return fileData
    }

    private suspend fun writeFile() {
        runSuspended(Dispatchers.IO) {
            val printWriter = file.printWriter()
            fileData?.forEach {
                printWriter.println("${it.key}=${it.value}")
            }
            printWriter.flush()
        }
    }

    override suspend fun delete(key: String): String? {
        val data = fileData ?: readFile()
        val value = data.remove(key)
        writeFile()
        return value
    }

    override fun entries(): Flow<RegistryStorage.StorageEntry<String, String>> {
        val data = fileData ?: readFile()
        return data.map { RegistryStorage.StorageEntry(it.key, it.value) }
            .asFlow()
    }

    override suspend fun read(key: String): String? {
        val data = fileData ?: readFile()
        return data[key]
    }

    override suspend fun upsert(key: String, value: String) {
        val data = fileData ?: readFile()
        data[key] = value
        writeFile()
    }

}
```

I think this is ready for a detailed review now!

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
```diff
@@ -14,52 +17,49 @@ import com.kotlindiscord.kord.extensions.commands.application.slash.SlashCommand
...
 import org.koin.core.component.inject
 import java.util.*
 
-/** Registry for all Discord application commands. **/
-public open class ApplicationCommandRegistry : KoinComponent {
-    private val logger = KotlinLogging.logger { }
+/**
+ * Abstraction layer between the [ExtensibleBot] and it's event handling.
+ *
+ * An implementing class is supposed register all commands and handle their discord events and lifecycles.
```
gdude2002: This isn't a fantastic doc comment, and it's not specific about what commands this handles - because, as I keep saying, KordEx has other types of commands.

Something along the following lines might be better:

```
Abstract class representing common behavior for application command registries.

Deals with the registration and syncing of, and dispatching to, all application commands.
Subtypes should build their functionality on top of this type.

@see DefaultApplicationCommandRegistry
```

Something like that.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
```diff
@@ -14,52 +17,49 @@ import com.kotlindiscord.kord.extensions.commands.application.slash.SlashCommand
...
+ */
+public abstract class ApplicationCommandRegistry : KoinComponent {
+
+    protected val logger: KLogger = KotlinLogging.logger { }
```
gdude2002, line 50: Make it `public` so that third-party subtypes can use it.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
```diff
@@ -72,324 +72,114 @@ public open class ApplicationCommandRegistry : KoinComponent {
...
-
-        try {
-            syncAll(true, commands)
-        } catch (t: Throwable) {
-            logger.error(t) { "Failed to synchronise application commands" }
-        }
-
+        initialize(commands)
         initialised = true
     }
```
gdude2002, line 98: You removed, like, all the newlines from this function. Put 'em back.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
```diff
@@ -72,324 +72,114 @@ public open class ApplicationCommandRegistry : KoinComponent {
...
             commands += it.userCommands
         }
-
-        try {
-            syncAll(true, commands)
-        } catch (t: Throwable) {
-            logger.error(t) { "Failed to synchronise application commands" }
-        }
-
+        initialize(commands)
```
gdude2002: Should probably wrap this with a `try` block to make sure `initialised` is always set.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
```diff
@@ -72,324 +72,114 @@ public open class ApplicationCommandRegistry : KoinComponent {
...
-                                commandObj.disallowedUsers.map { user(it, false) }
-
-                                commandObj.allowedRoles.map { role(it, true) }
-                                commandObj.disallowedRoles.map { role(it, false) }
-                            }
-                        }
-                    }
+    /** Register a [SlashCommand] to the registry.
+     *
+     * This method is only called after the [initialize] method and scopes to allow runtime modifications.
```
gdude2002: "Scopes to allow..."? What?

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
```diff
@@ -72,324 +72,114 @@ public open class ApplicationCommandRegistry : KoinComponent {
...
+    /** Unregister a user command. **/
+    public abstract suspend fun unregister(command: UserCommand<*>, delete: Boolean = true): UserCommand<*>?
 
-        // Extension commands that haven't been registered yet
-        val toAdd = commands.filter { c -> registered.all { !c.matches(locale, it) } }
+    // region: Utilities
 
-        // Extension commands that were previously registered
-        val toUpdate = commands.filter { c -> registered.any { c.matches(locale, it) } }
+    /** Unregister a message command. **/
```
gdude2002: This doc comment is wrong.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
```diff
@@ -72,324 +72,114 @@ public open class ApplicationCommandRegistry : KoinComponent {
...
-                        if (e.error?.message != null) {
-                            "\n        Discord error message: ${e.error?.message}"
-                        } else {
-                            ""
-                        }
+                when (it) {
+                    is SlashCommand<*, *> -> register(it) as T
+                    is MessageCommand<*> -> register(it) as T
+                    is UserCommand<*> -> register(it) as T
+                    else -> throw IllegalArgumentException(
```
gdude2002, line 191: Don't make this exhaustive, we _want_ the compiler to tell us when all branches aren't accounted for.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
```diff
@@ -406,14 +196,22 @@ public open class ApplicationCommandRegistry : KoinComponent {
...
 
-    // region: Typed registration functions
+    /**
+     * Creates a KordEx [ApplicationCommand] as discord command and returns the created command's id as [Snowflake].
+     */
+    public open suspend fun createDiscordCommand(command: ApplicationCommand<*>): Snowflake? = when (command) {
+        is SlashCommand<*, *> -> createDiscordSlashCommand(command)
+        is UserCommand<*> -> createDiscordUserCommand(command)
+        is MessageCommand<*> -> createDiscordMessageCommand(command)
+        else -> throw IllegalArgumentException("Unknown ApplicationCommand type")
```
gdude2002, line 221: Again, don't make this exhaustive. We want a compiler error here.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
```diff
@@ -423,63 +221,35 @@ public open class ApplicationCommandRegistry : KoinComponent {
...
+        injectPermissions(guild, command, response.id) ?: return null
 
-        return command
+        return response.id
     }
 
-    /** Register a slash command. **/
-    public open suspend fun register(command: SlashCommand<*, *>): SlashCommand<*, *>? {
+    /**
+     * Creates a KordEx [UserCommand] as discord command and returns the created command's id as [Snowflake].
```
gdude2002, line 263: Something like this would be more appropriate:

```
Register the given [UserCommand], syncing it to Discord.

@returns Snowflake representing the registered command's ID.
```

This is true for the other similar functions below.

Also, do these really need to have different names? The overloaded approach with the function named `register` was much clearer, in my opinion. If not, at least remove the `Discord` tautology and change it to, eg, `registerUserCommand`.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
```diff
@@ -560,29 +300,39 @@ public open class ApplicationCommandRegistry : KoinComponent {
...
+        injectPermissions(guild, command, response.id) ?: return null
+
+        return response.id
+    }
+
+    // endregion
+
+    // region: Permissions
+
+    protected suspend fun <T : ApplicationCommand<*>> injectPermissions(
```
gdude2002, line 340: Nope, this needs to be public for third-party subtypes.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
```diff
@@ -605,124 +355,18 @@ public open class ApplicationCommandRegistry : KoinComponent {
...
-        }
-
-        return userCommands.remove(id)
-    }
-
-    /** @suppress Internal function used to delete the given command from Discord. Used by [unregister]. **/
-    public open suspend fun deleteGeneric(
-        command: ApplicationCommand<*>,
-        discordCommandId: Snowflake,
+    protected fun injectRawPermissions(
```
gdude2002, line 374: This should also be open. In general, don't hide methods that will ever be useful in subtypes in any way - make them public, document them, and allow subtypes to use them.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
```diff
@@ -72,324 +72,114 @@ public open class ApplicationCommandRegistry : KoinComponent {
...
         initialised = true
     }
 
-    // region: Untyped sync functions
-
-    /** Register multiple generic application commands. **/
-    public open suspend fun syncAll(removeOthers: Boolean = false, commands: List<ApplicationCommand<*>>) {
-        val groupedCommands = commands.groupBy { it.guildId }
+    /** Called once the initial registration started and all extensions are loaded. **/
+    protected abstract suspend fun initialize(commands: List<ApplicationCommand<*>>)
```
gdude2002, line 101: Gotta be `public`.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/StorageAwareApplicationCommandRegistry.kt
```diff
@@ -0,0 +1,135 @@
...
+import dev.kord.common.entity.Snowflake
+import dev.kord.core.event.interaction.ChatInputCommandInteractionCreateEvent
+import dev.kord.core.event.interaction.MessageCommandInteractionCreateEvent
+import dev.kord.core.event.interaction.UserCommandInteractionCreateEvent
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.firstOrNull
+import kotlinx.coroutines.flow.toList
+
+/**
+ * ApplicationCommandRegistry which acts based off a specified storage interface.
```
gdude2002: `[]` please

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/StorageAwareApplicationCommandRegistry.kt
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * Discord lifecycles may not be implemented in this class and require manual updating.
```
gdude2002, line 25: Do you mean that this type may not handle syncing commands to Discord? That really should be something that's easy to do for users, so I don't really know whether that's ideal for this class.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/StorageAwareApplicationCommandRegistry.kt
```diff
@@ -0,0 +1,135 @@
...
+ */
+public class StorageAwareApplicationCommandRegistry(
```
gdude2002: I feel like it'd probably be useful for some people if this was `open` and also had `public` or `protected` access to everything you've restricted access to below - otherwise people will have to copy-paste the code in this class if they want to build upon it.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/StorageAwareApplicationCommandRegistry.kt
```diff
@@ -0,0 +1,135 @@
...
+    private val commandRegistry: RegistryStorage<Snowflake, ApplicationCommand<*>> = builder.invoke()
+
+    override suspend fun initialize(commands: List<ApplicationCommand<*>>) {
+        commands.forEach { commandRegistry.register(it) }
+
+        val registeredCommands = commandRegistry.entryFlow().toList()
+
+        commands.forEach { command ->
+            if (registeredCommands.none { it.hasCommand(command) }) {
+                val commandId = createDiscordCommand(command)
```
gdude2002, line 40: Let's get a blank line here.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/StorageAwareApplicationCommandRegistry.kt
```diff
@@ -0,0 +1,135 @@
...
+        unregisterApplicationCommand(command, delete) as? MessageCommand<*>
+
+    override suspend fun unregister(command: UserCommand<*>, delete: Boolean): UserCommand<*>? =
+        unregisterApplicationCommand(command, delete) as? UserCommand<*>
+
+    private suspend fun unregisterApplicationCommand(
+        command: ApplicationCommand<*>,
+        delete: Boolean
+    ): ApplicationCommand<*>? {
+        val id = commandRegistry.constructUniqueIdentifier(command)
```
gdude2002, line 113: Let's get a blank line here as well.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/StorageAwareApplicationCommandRegistry.kt
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        return null
+    }
+
+    private fun RegistryStorage.StorageEntry<Snowflake, ApplicationCommand<*>>.hasCommand(
+        command: ApplicationCommand<*>
+    ): Boolean {
+        val key = commandRegistry.constructUniqueIdentifier(value)
+        val other = commandRegistry.constructUniqueIdentifier(command)
```
gdude2002, line 134: And one here, please.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/registry/AbstractDeconstructingApplicationCommandRegistryStorage.kt
```diff
@@ -0,0 +1,79 @@
...
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.mapNotNull
+
+/**
+ * Abstract class which can be used to implement a simple networked key-value based registry storage
+ * for [ApplicationCommand]s.
+ *
+ * For simplicity the parameter / return types of the abstract methods are all [String]s.
+ */
+public abstract class AbstractDeconstructingApplicationCommandRegistryStorage<T : ApplicationCommand<*>> :
```
gdude2002, line 14: You've got an abstract class here with a bunch of private stuff. This all needs to be `public` or `protected`, this is a framework we've got here.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/registry/DefaultLocalRegistryStorage.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.asFlow
+import kotlinx.coroutines.flow.map
+
+/**
+ * Default "local" implementation of [RegistryStorage] which internally
+ * uses a mutableMap.
+ */
+public class DefaultLocalRegistryStorage<K, T> : RegistryStorage<K, T> {
```
gdude2002: Make it `open`, make things `public` or `protected`. You know the drill. I'd stick a `@suppress` in the `registry` kdoc, though.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/registry/RegistryStorage.kt
```diff
@@ -0,0 +1,65 @@
...
+ * Interface for interaction based registries like ComponentRegistry and ApplicationCommandRegistry.
+ *
+ * The purpose of this interface is to provide a generic way to store Components/ApplicationCommands
+ * in a dynamic manner.
+ */
+public interface RegistryStorage<K, T> {
+
+    /**
+     * Let's the registry know about the specified type, this may store the object in a local map,
+     * which is used for reconstructing later.
```
gdude2002, line 15: This needs a bit more detail, IMO. Also, it's `lets` in this context.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/registry/RegistryStorage.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Constructs a unique key for the given data.
+     */
+    public fun constructUniqueIdentifier(data: T): String
+
+    /**
+     * Data class to represent an entry in the [RegistryStorage].
+     */
+    public data class StorageEntry<K, V>(
```
gdude2002, line 55: This is basically just a `Pair`. Any reason why you didn't use that?

gdude2002 (changes requested):
Good to see your hard work here. Just a few things for you to look at!

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
ByteAlex: It is

gdude2002: Yup yup, visibility modifier confusion as mentioned on Discord

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/StorageAwareApplicationCommandRegistry.kt
ByteAlex: It only does partial updates, i.e. only registers new, but doesnt check for updates nor deletions.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/StorageAwareApplicationCommandRegistry.kt
ByteAlex: Sure, this should be an easy fix.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/StorageAwareApplicationCommandRegistry.kt
gdude2002: If that's the case, then we really ought to have an advanced tutorial that explains how to do this in a nice way.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/registry/RegistryStorage.kt
ByteAlex: I don't like the .0 .1 syntax on `Pair` and overall this explain in which scope this "Pair" lives.
Also I was playing with putting the registry into the entry as an additional reference, yet decided against it. This would also allow for further extension of this without breaking other's code.

gdude2002: Hmm, fair point.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
ByteAlex: Heh? Why that?

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
ByteAlex: It needs to be exhaustive, it won't compile without it.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
ByteAlex: It needs to be exhaustive, it won't compile without it.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
gdude2002: It's a generic unregistration function, not just for message commands.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
gdude2002: Hmm, maybe I need to seal `ApplicationCommand`? Think that'd help?

Or, wait, then subtypes have to be in the same file, right? That's not great.

I guess if there's nothing we can do, we'll have to stick with it. But you can put a newline before the `else` branch. :>

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
ByteAlex: No, it actually does exactly what the name says. It `creates` a `Discord` command of the given type. It does not do *anything* with the registry. This is solely used to create the commands on discord's backend.

gdude2002: That doesn't make sense to me, API-wise - if you're going to tell Discord about the command then you're going to want to handle it as well, so it doesn't seem sensible for this to not touch the registry.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
ByteAlex: Oh lol. I looked at the wrong method, haha.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
ByteAlex: Fine by me

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
ByteAlex: Previously you had this somewhere in your code and it was duplicated code, I just moved it to a designated function. And it still makes sense in that scope, becaue most subtypes would want to register application commands in Discord's backend at some point.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/StorageAwareApplicationCommandRegistry.kt
ByteAlex: That will probably make sense, though if someone really needs this, they'll probably know what they're doing already... I hope?

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/ApplicationCommandRegistry.kt
gdude2002: Ah, okay, I see what you mean.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/StorageAwareApplicationCommandRegistry.kt
gdude2002: I guess so, though there's no reason we can't put something on the wiki later.

gdude2002 (approved):
:rocket:

Issue #90, opened 2021-09-25
Lazy loading on Paginators


ByteAlex:
# Summary

Paginators should support lazy loading pages with either "infinite scroll" or a fixed ending.

# Current Behaviour

Pages need to be predefined from the creation of the message.

# Preferred Behaviour

Pages should be lazy loaded as a user may click through them.

# Reasoning

A database may be large; for example a leaderboard. You need a sorted subset of it, but not all data at once. That saves a lot of traffic and cpu cycles for data which most likely is never been seen anyway... Who is gonna click the "next page" 25 times every time they open the leaderboard, right?

PR #100, opened 2021-10-03
Add convenience configurator for collection paginators

DRSchlaubi:
Currently there isn't any easy way to cerate a paginator just for a list of items, this PR aims to fix that﻿

### DIFF
--- a/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/pagination/CollectionPaginator.kt
+++ b/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/pagination/CollectionPaginator.kt
@@ -1,0 +1,107 @@
+package com.kotlindiscord.kord.extensions.pagination
+
+import com.kotlindiscord.kord.extensions.pagination.builders.PaginatorBuilder
+import dev.kord.rest.builder.message.EmbedBuilder
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.drop
+import kotlinx.coroutines.flow.take
+import kotlinx.coroutines.flow.toList
+import kotlin.math.ceil
+
+/**
+ * Configures this [PaginatorBuilder] to create one page for each [x][pageSize] elements in [items].
+ *
+ * @param items a [List] containing all the items
+ * @param mapper a mapper converting [T] to [String]
+ * @param title a function providing the title for the current page
+ * @param enumerate whether to include element numbers in entries or not
+ * @param additionalPageConfig additional [EmbedBuilder] config, applied to each page
+ */
+public suspend fun <T> PaginatorBuilder.forList(
+    items: List<T>,
+    mapper: suspend (T) -> String,
+    title: suspend (current: Int, total: Int) -> String,
+    pageSize: Int = 8,
+    enumerate: Boolean = true,
+    additionalPageConfig: suspend EmbedBuilder.() -> Unit = {}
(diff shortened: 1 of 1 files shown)


gdude2002:
This API seems far too complex and domain-specific. The functions are, in general, pretty nuts - this is something that requires too much mental overhead to include as a user-facing API.

What's your use-case for this? Why not just `collection.forEach { page { ... } }`?

DRSchlaubi:
> This API seems far too complex and domain-specific. The functions are, in general, pretty nuts - this is something that requires too much mental overhead to include as a user-facing API.

This seems like the most common use case for paginators, just listing a collection

> What's your use-case for this? Why not just `collection.forEach { page { ... } }`
This functions also allow usage of e.g. Flows and request content on the fly instead of requesting all the pages at once, also always having to do the manual formatting is pretty annoying.

Also, this is pretty similar to what JDA utilities does here: https://github.com/JDA-Applications/JDA-Utilities/blob/master/menu/src/main/java/com/jagrosh/jdautilities/menu/Paginator.java



<blockquote><div> GitHub</div><div><strong><a href="https://github.com/JDA-Applications/JDA-Utilities">JDA-Utilities/Paginator.java at master · JDA-Applications/JDA-Utilities</a></strong></div><div>A series of tools and utilities for JDA to assist in bot creation - JDA-Utilities/Paginator.java at master · JDA-Applications/JDA-Utilities</div></blockquote>

gdude2002:
> This seems like the most common use case for paginators, just listing a collection

Right, but you're not just listing a collection here, you're importing a collection with 4 separate required arguments most of which don't seem like they'd be useful for most people.

> This functions also allow usage of e.g. Flows and request content on the fly instead of requesting all the pages at once, also always having to do the manual formatting is pretty annoying.

I have different ideas for lazy paginators that I'll be exploring in #90 - I don't think just throwing a flow into the builder is going to help much.

> Also, this is pretty similar to what JDA utilities does here [ ... ]

Eh, where? This is a 730 line file.

DRSchlaubi:
> Right, but you're not just listing a collection here, you're importing a collection with 4 separate required arguments most of which don't seem like they'd be useful for most people.

There are 6 of them, but only 3 of them are actually required, and we could eliminate one of them if we add a function specifically for strings, and we could eliminate another one if we specify a default title

> I have different ideas for lazy paginators that I'll be exploring in #90 - I don't think just throwing a flow into the builder is going to help much.

A page stores the builder as this `suspend EmbedBuilder.() -> Unit` so the request function of the paginator gets called when rendering the page, which means that depending on the flow implementation this is lazy

gdude2002:
> A page stores the builder as this `suspend EmbedBuilder.() -> Unit` so the request function of the paginator gets called when rendering the page, which means that depending on the flow implementation this is lazy

That's really only technically lazy, not lazy in a meaningful way. Either way, the Flow has to be exhausted for the paginator to be shown - that's not ideal for several use-cases that I can think of.

> There are 6 of them, but only 3 of them are actually required, and we could eliminate one of them if we add a function specifically for strings, and we could eliminate another one if we specify a default title

Pages aren't... Strings, though? They're full embed builders. That was a deliberate choice for formatting flexibility, and it's been a suuuuuuper long time since pages were restricted this way.

gdude2002:
Marking this as stale as it hasn't had activity for well over a week.

gdude2002:
This has been stale for too long, so I'm closing it for now. I already had plans for this feature, so don't worry too much - it's just one of those things.

PR #102, opened 2021-10-09
Add converter for discord-formatted timestamps

### LINKED ISSUE #85: Converters that handle discord formatted timestamps
Title

Scotsguy:
Closes #85 

Not bothering with the other time modules for now


gdude2002:
Thanks, I'm happy with this addition. It doesn't quite close #85 as I also intended for this support to be added to the duration converters (for the sake of completeness), but this PR can be merged without that if you'd prefer.

Scotsguy:
How would you expect that to look like? Discord timestamps are, well, timestamps, not durations.

gdude2002:
They're not durations, but they do point to a specific point in time - so it'd be a case of getting the delta between "now" and the timestamp as a duration

gdude2002:
This looks good overall. Thanks!

🚀

PR #105, opened 2021-11-08
Make ResourceBundleTranslations

DRSchlaubi:
Make ResourceBundleTranslations open, can be useful if you want to customize behavior of resource bundle loading


DRSchlaubi:
ofc my history is completly broken

PR #109, opened 2021-11-16
Mappings to Slash Commands

sschr15:
- Migrate mapping-related commands to use slash commands
- Add slash commands for mapping conversions


gdude2002:
This pull request is to the wrong branch - as I said when you asked me on Discord, you need to PR to `develop`.

PR #112, opened 2021-11-16
More method conversion fixes

sschr15:
detekt is helpful but it made me write code that I'd consider uglier just to not go over that whole maximum line length restriction


gdude2002:
Third time lucky, eh? Good work regardless, though

PR #113, opened 2021-11-22
Mapping fixes again

sschr15:


sschr15:
apparently merge conflicts? aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

gdude2002:
Instead of relying on a gist that's never going to need to be updated and may disappear, maybe we should just include the JSON file as a resource? Or is that not possible?

gdude2002:
🚀 

Thank

PR #114, opened 2021-11-22
make code into less code which is functionally equivalent

sschr15:
unless i did something terribly wrong, the code *should* give the same results to the end user as if i made no changes


gdude2002:
this is a lot of PRs, lmao

PR #118, opened 2021-12-03
Add an optional map descriptors option for mappings

sschr15:
this time with an up-to-date upstream

### DIFF
--- a/extra-modules/extra-mappings/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/mappings/arguments/MCPArguments.kt
+++ b/extra-modules/extra-mappings/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/mappings/arguments/MCPArguments.kt
@@ -4,4 +4,7 @@ import com.kotlindiscord.kord.extensions.modules.extra.mappings.utils.McpNamespa
 
 /** Arguments for MCP mappings lookup commands. **/
 @Suppress("UndocumentedPublicProperty")
-class MCPArguments : MappingArguments(McpNamespaceReplacement)
+class MCPArguments : MappingArguments(McpNamespaceReplacement) {
+    // MCP doesn't support intermediary descriptors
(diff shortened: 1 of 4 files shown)


extra-modules/extra-mappings/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/mappings/arguments/MappingArguments.kt
```diff
@@ -18,4 +19,10 @@ open class MappingArguments(val namespace: Namespace) : Arguments() {
         true,
         namespace
     )
+
+    open val mapDescriptors by defaultingBoolean(
+        "mapDescriptors",
```
gdude2002, line 24: This is user-facing, it shouldn't use programmer casing. `kebab-case` is standard.

extra-modules/extra-mappings/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/mappings/arguments/MCPArguments.kt
```diff
@@ -4,4 +4,7 @@ import com.kotlindiscord.kord.extensions.modules.extra.mappings.utils.McpNamespa
 
 /** Arguments for MCP mappings lookup commands. **/
 @Suppress("UndocumentedPublicProperty")
-class MCPArguments : MappingArguments(McpNamespaceReplacement)
+class MCPArguments : MappingArguments(McpNamespaceReplacement) {
+    // MCP doesn't support intermediary descriptors
+    override val mapDescriptors = true
```
gdude2002, line 9: Where's this used? We should document which mappings commands don't support supplying this, and set up the commands for them to not take this argument.

gdude2002 (changes requested):

sschr15:
Better version and update was pulled in #127

PR #128, opened 2022-01-17
Add auto complete

DRSchlaubi:


gdude2002:
Thanks for your work on this - ultimately I decided to implement it a little differently, esp. as you said yesterday that AC is broken anyway.

Still, thanks for looking into it!

PR #131, opened 2022-01-26
Add Markdown Sanitizer functionality

DRSchlaubi:
JDA had these functions as well so might be helpful here too


gdude2002:
Is this ported from JDA?

gdude2002:
It looks like it's ported from JDA.

DRSchlaubi:
The only thing ported from JDA is the docs string on [contentDisplay]

DRSchlaubi:
Superseded by kordlib/kord#521

PR #138, opened 2022-02-25
Update to Kord M10

lukellmann:
Kord 0.8.0-M10 included a major refactor with breaking changes, this PR updates kordex to work with the latest snapshots by updating imports, types and fixing deprecations.

~However this will not work with 0.8.0-M10 directly because of a small oversight (missing `command` in `AutoCompleteInteraction`) that was fixed right after M10.~
I added a drop in replacement for the missing `command` property.

PR #139, opened 2022-03-02
feat: warn on duplicate extension name

Jamalam360:
Very small change but PRing because I know you've said you've been busy, gdude.


Jamalam360:
I don't _think_ those actions errors are my fault. Builds fine locally, and I can't see why this would cause those errors

lukellmann:
> I don't _think_ those actions errors are my fault. Builds fine locally, and I can't see why this would cause those errors

It's because of Kord's snapshot version in the develop branch, there was a breaking change (the reason for https://github.com/Kord-Extensions/kord-extensions/pull/138)

Jamalam360:
Oh ok that's fine then

gdude2002:
#138 has been merged, so you'll want to rebase

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/ExtensibleBot.kt
```diff
@@ -266,6 +266,10 @@ public open class ExtensibleBot(
     public open suspend fun addExtension(builder: () -> Extension) {
         val extensionObj = builder.invoke()
 
+        if (extensions.contains(extensionObj.name)) {
+            logger.warn { "Using duplicate extension name: ${extensionObj.name}" }
```
gdude2002: This should use `logger.error` instead and unload the old extension first - duplicate extension names are _breaking_, as the already existing extension will no longer be able to be referred to.

gdude2002 (changes requested):

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/ExtensibleBot.kt
gdude2002: In case you only look at emails: Edited the original comment

Jamalam360: Yeah OK, I wasn't sure if you'd want the default behaviour changing, pushing now

gdude2002: The push is good, but the message isn't clear enough. Maybe something like...

`Extension with duplicate name ${extensionObj.name} registered - unloading previously registered extension.`

gdude2002 (approved):

gdude2002:
Thank!

Jamalam360:
:)

PR #140, opened 2022-03-12
Add repeatable tasks

sschr15:
this was actually a bit easier than i had imagined, i just had to remember to do things properly

also there's a small change just to get rid of a nullable `Unit` which surprisingly greatly increases the compiled result even though performance or compilation optimizations are likely not high on the list of importance but i did it anyway

### DIFF
--- a/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/scheduling/Task.kt
+++ b/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/scheduling/Task.kt
@@ -126,22 +126,22 @@ public open class Task(
         } catch (t: Throwable) {
             logger.error(t) { "Error running scheduled callback." }
         }
-
-        if (!repeat) {
-            removeFromParent()
-        }
     }
 
     /** Stop waiting and don't execute. **/
     public fun cancel() {
         job?.cancel()
         job = null
+
+        removeFromParent()
     }
 
     /** Like [cancel], but blocks .. **/
     public suspend fun cancelAndJoin() {
         job?.cancelAndJoin()
         job = null
+
+        removeFromParent()
     }
 
     /** If the task is running, cancel it and restart it. **/
--- a/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/scheduling/Task.kt
+++ b/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/scheduling/Task.kt
@@ -35,6 +35,7 @@ import kotlin.time.TimeSource
  * @param coroutineScope Coroutine scope to launch in - Kord's by default.
  * @param parent Parent [Scheduler] object, if any.
  * @param name Optional task name, "Unnamed" by default.
+ * @param repeat Whether the task should repeat after completion. `false` by default.
  */
 public open class Task(
     public open val duration: Duration,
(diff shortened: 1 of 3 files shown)


kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/scheduling/Task.kt
```diff
@@ -104,9 +106,19 @@ public open class Task(
...
 
-            removeFromParent()
+            if (!repeat) {
+                removeFromParent()
+            }
 
             job = null
+
+            if (repeat) {
+                sentryContext.breadcrumb(BreadcrumbType.Info) {
```
gdude2002: I think this could result in a memory leak, considering you add a breadcrumb every time the task repeats. I think it'd probably be simplest to just remove it, but I'm open to better ideas.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/scheduling/Task.kt
```diff
@@ -121,23 +133,29 @@ public open class Task(
             logger.error(t) { "Error running scheduled callback." }
         }
 
-        removeFromParent()
+        if (!repeat) {
```
gdude2002: This can be removed as `cancel()` already does it.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/scheduling/Task.kt
```diff
@@ -121,23 +133,29 @@ public open class Task(
...
 
     /** Stop waiting and don't execute. **/
     public fun cancel() {
         job?.cancel()
         job = null
 
-        removeFromParent()
+        if (!repeat) {
+            removeFromParent()
+        }
```
gdude2002: There's no point in adding this conditional here, it should _always_ be removed as it stops running at this point.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/scheduling/Task.kt
```diff
@@ -121,23 +133,29 @@ public open class Task(
...
 
     /** Like [cancel], but blocks .. **/
     public suspend fun cancelAndJoin() {
         job?.cancelAndJoin()
         job = null
 
-        removeFromParent()
+        if (!repeat) {
+            removeFromParent()
+        }
```
gdude2002: There's no point in adding this conditional here, it should _always_ be removed as it stops running at this point.

gdude2002 (changes requested):

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/scheduling/Task.kt
sschr15: This actually can't be removed because if `callNow()` is called, the task won't remove itself ever (partially due to no cancellation checks)

gdude2002: `callNow()` calls `cancel()` though, which does the removal.

gdude2002:
This looks good to me - thanks for doing it!

PR #145, opened 2022-04-15
Fix inconsistent/extension plugin loading

DSeeLP:
Coroutines were started in Plugin start/stop/delete but weren't joined which lead to inconsistent loading of plugins and extensions.
I also replaced hard coded test server ids in TestExtension with the TEST_SERVER_ID constant.


gdude2002:
Looks good to me - thanks!

PR #146, opened 2022-04-16
Cleaning: Abstractions, old docs or docs mistake fixes

ToxicMushroom:
Motivation: abstracting duplicate code leads to less code to maintain, less chance to miss one place.
I also added some quick documentation/type fixes that I found along the way.
I'll try to put these in separate pr's in the future :>


gdude2002:
Looks good to me - will squash in after the build passes.

gdude2002:
Thanks o7

PR #154, opened 2022-05-09
Adding support for bot restarts

nicholasgrose:
This adds a stop() function to the extensible bot. In order to support it, it also adds an improved implementation of setupKoin() in the ExtensibleBotBuilder. This implementation checks that a Koin instance exists before attempting to start it, configuring the instance afterwards. It also adds a list of bot-related Koin modules that is added to and unloaded when the bot is stopped.


gdude2002:
This really doesn't feel like the way to go at all - we'd be better off with a KordEx-managed Koin instance instead, although I'm not actually sure how that would need to be implemented. Messing with module load order like this is likely to introduce subtle bugs, and removing Koin modules like this simply isn't going to work in most use-cases.

nicholasgrose:
> This really doesn't feel like the way to go at all - we'd be better off with a KordEx-managed Koin instance instead, although I'm not actually sure how that would need to be implemented. Messing with module load order like this is likely to introduce subtle bugs, and removing Koin modules like this simply isn't going to work in most use-cases.

That's understandable, it does feel a little jerry-rigged, in retrospect. However, I have found a far more elegant solution, specifically recommended by Koin.

I found [this article in the Koin docs](https://insert-koin.io/docs/reference/koin-core/context-isolation), which references context isolation. It seems that context isolation specifically exists for libraries to solve the problem of overlapping Koin contexts. I could implement it like that, instead, stopping that application on bot stopping?

gdude2002:
That's not a bad approach, but I think it needs to be done in a way that doesn't prevent multiple KordEx bots from existing together. Unfortunately, though, when I did that before, it basically defeated the point of using Koin in the first place - because I had to pass a container type around everywhere.

The `object` approach is a good way to stop things from conflicting, though.

nicholasgrose:
> That's not a bad approach, but I think it needs to be done in a way that doesn't prevent multiple KordEx bots from existing together. Unfortunately, though, when I did that before, it basically defeated the point of using Koin in the first place - because I had to pass a container type around everywhere.
> 
> The `object` approach is a good way to stop things from conflicting, though.

Isn't that already being prevented, though? If you start up a bot on top of another bot, the global Koin context will overlap and an error will be thrown before you can start up the second bot.

To implement that, it could be possible to, rather than having a single global instance, keep a global map of bots to their respective instances. Thus, they can be isolated from each other, and we could keep a consistent implementation for getting the Koin instance for any particular bot.

gdude2002:
But now you have to pass around an instance of the bot instead?

nicholasgrose:
> But now you have to pass around an instance of the bot instead?

Ah. True. I'm not certain of any method besides just keeping an `object` and using that for the KoinApplication.

It feels like [scopes](https://insert-koin.io/docs/reference/koin-core/scopes#what-is-a-scope) should somehow be involved with a solution allowing multiple bots, but that still requires that the scope get passed around everywhere.

nicholasgrose:
Since multiple bots is currently not possible, would it be reasonable to implement the KoinApplication as being contained in an `object` for the time being? It would not allow multiple bots, but, since that is not currently possible as it is, it would not actually be significantly changing functionality.

gdude2002:
Yeah it would, I think I already said that haha

nicholasgrose:
> Yeah it would, I think I already said that haha

I suppose I misunderstood, then! Ha ha. I'll make that change.

nicholasgrose:
I've now added an isolated Koin context, and I've removed the old module unloading code. Due to the number of places KoinComponent is used, I wound up having a lot of files changed, but all those changes are minor. Let me know if I missed any locations that should be using KordExKoinComponent or KordExContext that are not.

I've now also added a function called `close()` on the extensible bot in addition to `stop()`. I did this because I realized that calling stop() made it sound as if the bot could be restarted when it could not be, since the builder is the one that sets up Koin dependencies. Thus, `stop()` will only shutdown Kord, which should allow for subsequent calls to `start()`. Similarly, `close()` has documentation that now says that restarting the same bot after its invocation will result in undefined behavior, since the Koin instance no longer exists.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/builders/ExtensibleBotBuilder.kt
```diff
@@ -345,38 +345,53 @@ public open class ExtensibleBotBuilder {
...
+
+    /** @suppress Internal function that creates and loads the bot's main Koin modules.
+     * The modules provide important bot-related singletons. **/
+    private fun addBotKoinModule() {
         loadModule { single { this@ExtensibleBotBuilder } bind ExtensibleBotBuilder::class }
         loadModule { single { i18nBuilder.translationsProvider } bind TranslationsProvider::class }
         loadModule { single { chatCommandsBuilder.registryBuilder() } bind ChatCommandRegistry::class }
         loadModule { single { componentsBuilder.registryBuilder() } bind ComponentRegistry::class }
         loadModule { single { componentsBuilder.callbackRegistryBuilder() } bind ComponentCallbackRegistry::class }
-
```
gdude2002: You've removed some newlines here - please don't do that, they're important for readability.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/ExtensibleBot.kt
```diff
@@ -126,6 +127,17 @@ public open class ExtensibleBot(
         }
     }
 
+    /** Stop the bot. **/
+    public open suspend fun stop() {
+        getKoin().get<Kord>().shutdown()
+    }
+
+    /** Stop the bot and its [Kord] instance. Restarting the bot after closing will result in undefined behavior. **/
```
gdude2002: This could be a bit more descriptive - it should explain _why_ undefined behaviour may occur (explain about the Koin closure).

That said, I think actually what'll happen is it'll just crash immediately, since the Koin context has been destroyed.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/builders/ExtensibleBotBuilder.kt
```diff
@@ -345,38 +345,53 @@ public open class ExtensibleBotBuilder {
...
+        startKoinIfNeeded()
 
-            if (logLevel == Level.INFO || logLevel == Level.DEBUG) {
-                // NOTE: Temporary workaround for Koin not supporting Kotlin 1.6
-                logLevel = Level.ERROR
-            }
+        hooksBuilder.runBeforeKoinSetup()
 
-            slf4jLogger(logLevel)
-//            environmentProperties()  // https://github.com/InsertKoinIO/koin/issues/1099
```
gdude2002, line 356: Can you bring this line over - it's a useful reminder

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/ExtensibleBot.kt
```diff
@@ -126,6 +127,17 @@ public open class ExtensibleBot(
         }
     }
 
+    /** Stop the bot. **/
```
gdude2002: Probably good to add something like "by shutting down Kord".

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/koin/KordExContext.kt
```diff
@@ -0,0 +1,125 @@
...
+ * To use this context, implement [KordExKoinComponent].
+ */
+public object KordExContext : KoinContext {
+    /** The current [Koin] instance. */
+    private var koin: Koin? = null
+
+    /** The current [KoinApplication]. */
+    private var koinApp: KoinApplication? = null
+
+    /** Gets the [Koin] instance.
```
gdude2002: Put that on the next line, please, as is the convention in this project. You can assume this for the rest of the multi-line block comments in this file as well.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/koin/KordExContext.kt
```diff
@@ -0,0 +1,125 @@
...
+    /** Registers a [KoinApplication] to as the current one for this context.
+     *
+     * @param koinApplication The application to registers.
+     *
+     * @throws KoinAppAlreadyStartedException The [KoinApplication] has already been instantiated.
+     */
+    private fun register(koinApplication: KoinApplication) {
+        if (koin != null) {
+            throw KoinAppAlreadyStartedException("KordEx Koin Application has already been started")
+        }
```
gdude2002, line 55: Could do with a blank line after this curly.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/koin/KordExContext.kt
```diff
@@ -0,0 +1,125 @@
...
+     * Starts using the provided [KoinApplication] as the current one for this context.
+     *
+     * @param koinApplication The application to start with.
+     *
+     * @throws KoinAppAlreadyStartedException The [KoinApplication] has already been instantiated.
+     */
+    override fun startKoin(koinApplication: KoinApplication): KoinApplication = synchronized(this) {
+        register(koinApplication)
+        koinApplication.createEagerInstances()
+        return koinApplication
```
gdude2002, line 78: Let's stick a blank line here for readability.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/koin/KordExContext.kt
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * Starts using the provided [KoinAppDeclaration] to create the [KoinApplication] for this context.
+     *
+     * @param appDeclaration The application declaration to start with.
+     *
+     * @throws KoinAppAlreadyStartedException The [KoinApplication] has already been instantiated.
+     */
+    override fun startKoin(appDeclaration: KoinAppDeclaration): KoinApplication = synchronized(this) {
+        val koinApplication = KoinApplication.init()
```
gdude2002, line 89: Let's stick a blank line here for readability as well.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/koin/KordExContext.kt
```diff
@@ -0,0 +1,125 @@
...
+        register(koinApplication)
+        appDeclaration(koinApplication)
+        koinApplication.createEagerInstances()
+        return koinApplication
```
gdude2002, line 95: You're getting the idea behind this project's spacing now, right? Haha

gdude2002 (changes requested):
Looks pretty good other than a few small things - have you tested this by running the test bot?

nicholasgrose:
Latest commit fixes all requested changes and even adds a few more, in terms of improving documentation of the code. I have not yet run the test bot, at the time of writing, but I'll have a look around the code base to see if I can figure out how to do so.

nicholasgrose:
I've found the test bot, and I've run some tests. I can confirm that the stopping functionality is working, however it appears that Kord is not appreciating getting shutdown and then restarted, claiming that "The Gateway has been detached and can no longer be used, create a new instance instead.". I'll do some testing to solve that issue.

nicholasgrose:
Alright, the restarting issue got ironed out. I switched `stop()` to `logout()`, rather than `shutdown()`. `close()` will now run `shutdown()` in its stead.

I also, in my testing, encountered an issue with double registration of event subscribers. I've fixed this by setting `registerListeners()`, in `start()`, to only run if it is not initialized (that is, it has not been run before). This allows for restarting without issue.

gdude2002 (approved):

gdude2002:
This PR is looking good, but it doesn't build in CI - please run a clean build and check the Actions logs, and fix that.

nicholasgrose:
I must say that I'm glad this error occurred. I found that I had apparently added the new KordExKoinComponent wrongly to a class, and it made me realize that that should be included everywhere, not just in kord-extensions. I'm assuming that anywhere we are using KoinComponent, I should swap it, right? Especially in 'modules/', since those are looking for Koin modules that are loaded from the ExtensibleBotBuilder.

nicholasgrose:
After swapping out `KoinComponent` for `KordExKoinComponent` in all locations that have access to it, I noticed that the only other place using Koin is within annotation-processor. That appears to have its own Koin context going on, so I'll leave it be. It may be a good thing to look at later to switch that to using a similarly isolated `KoinContext`.

gdude2002:
Yeah, the annotation processor doesn't need that, it doesn't run at runtime!

This looks good, good work \o/

gdude2002:
I'm going to hold off on merging this just until #155 has been merged and Kord has been updated - since this is a breaking change, it needs a version bump!

nicholasgrose:
Sounds good! Thank you, and thank you for your time!

nicholasgrose:
Latest commit fixes merge conflicts created by the most recent commits. I tested the stopping functionality again with the test bot, and it is still functioning as expected.

gdude2002:
Great work, thanks for this o7

Issue #158, opened 2022-05-21
Throws exception, when using translated slash subcommands


zxcqirara:
# Description

I registered slash **sub**command (eg `/remind list`), but when I used it, bot throwed exception (with translated slash commands works fine)

Exception log:
```
Exception in thread "DefaultDispatcher-worker-12" java.lang.IllegalStateException: Unknown subcommand: list
    at com.kotlindiscord.kord.extensions.commands.application.slash.SlashCommand.findCommand(SlashCommand.kt:211)
    at com.kotlindiscord.kord.extensions.commands.application.slash.SlashCommand.findCommand(SlashCommand.kt:198)
    at com.kotlindiscord.kord.extensions.commands.application.slash.EphemeralSlashCommand.call(EphemeralSlashCommand.kt:44)
    at com.kotlindiscord.kord.extensions.commands.application.DefaultApplicationCommandRegistry.handle$suspendImpl(DefaultApplicationCommandRegistry.kt:353)
    at com.kotlindiscord.kord.extensions.commands.application.DefaultApplicationCommandRegistry.handle(DefaultApplicationCommandRegistry.kt)
    at com.kotlindiscord.kord.extensions.ExtensibleBot$registerListeners$8.invokeSuspend(ExtensibleBot.kt:177)
    at com.kotlindiscord.kord.extensions.ExtensibleBot$registerListeners$8.invoke(ExtensibleBot.kt)
    at com.kotlindiscord.kord.extensions.ExtensibleBot$registerListeners$8.invoke(ExtensibleBot.kt)
    at com.kotlindiscord.kord.extensions.ExtensibleBot$registerListeners$suspendImpl$$inlined$on$default$17$1.invokeSuspend(ExtensibleBot.kt:243)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
    at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
    at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:570)
    at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:749)
    at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:677)
[...]

PR #162, opened 2022-07-07
Fix localized slash commands & groups

NyCodeGHG:
# Relevant Issues
#158

# Description

Discord is not able to handle localization keys as subcommand or group names when using localized subcommands/groups.
So now we just send them the default translation for the name instead of the translation key beside the translations.

I also updated the test bot to not have the same translation keys as translations because this caused to not be able to reproduce the bug described in #158


gdude2002:
As mentioned in #158, this seems inaccurate --- or at least it doesn't match Discord's intended behaviour. It's worth noting that slash command localization isn't ready for production even by Discord's standards, so we'll need to chat with them a bit more about this I think.

gdude2002:
After some discussion on Discord and further review, it seems that for whatever reason we had a miscommunication - this PR looks good to me, thanks for your hard work!

PR #163, opened 2022-07-07
Open up converters

NyCodeGHG:
develop

### DIFF
--- a/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/converters/CoalescingToDefaultingConverter.kt
+++ b/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/converters/CoalescingToDefaultingConverter.kt
@@ -25,7 +25,7 @@ import dev.kord.rest.builder.interaction.OptionsBuilder
  * [coalescingConverter].
  * @param newErrorTypeString An optional error type string to override the one set in [coalescingConverter].
  */
-public class CoalescingToDefaultingConverter<T : Any>(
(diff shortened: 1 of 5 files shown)


NyCodeGHG:
oops i guess that went a bit wrong

NyCodeGHG:
ok, for whatever reason github's cli did not let me fill out the pr template.
I need do modify some stuff in a converter, but i can't extend it because it's not open, thats it basically.

gdude2002:
Can you explain what your use-case is here? I'd be curious to know whether there's a better option first

NyCodeGHG:
I have [this converter](https://github.com/mikbot/hafalsch/blob/a00f5416d639c52621cec79b00bccef0d4a02a2a/bot/src/main/kotlin/dev/schlaubi/hafalsch/bot/command/AutoCompletingArgument.kt#L18-L21) but i need to override toOptional and provide my own subclass of SingleToOptionalConverter and need to also set the autocomplete lambda in the builder there. But I just realized that it would be smarter to make sure SingleToOptionalConverter uses the same builder options

gdude2002:
It's worth noting that everything is type-safe here, including the builders themselves, so you could reasonably take a generic ConverterBuilder type here - but yeah, I'm not sure to what extent that helps you.

We could open these, I guess, but it's worth noting that these "converter converters" are really only meant for automatically wrapping other converters. But either way, let's chat on Discord a little so we can see where we're at.

NyCodeGHG:
Closing because use case can be done otherwise

PR #164, opened 2022-07-08
Fix printVersion task

NyCodeGHG:
# Description

I noticed the printVersion task is defined wrong. Its content gets executed everytime the buildscript gets ran.
Now it just prints when the task gets executed.


gdude2002:
Thanks!

PR #166, opened 2022-07-10
Fix option registration

NyCodeGHG:
# Description
Slash command options were registered with their default translation, but then compared to their translation key.
This behavior broke option autocomplete. See media. This PR fixes that by comparing to the default translation.

# Media
## Before
[Peek 2022-07-11 01-35.webm](https://user-images.githubusercontent.com/37078297/178166725-741ac924-a2f9-4eaf-8417-0add1a871a0c.webm)

## After
[Peek 2022-07-11 01-47.webm](https://user-images.githubusercontent.com/37078297/178166747-db6bd567-89ab-4b46-a033-ed9729a89fe5.webm)


gdude2002:
Another high quality PR. Thank you!

PR #172, opened 2022-08-07
Add Manage Webhooks permission requirement to PK Module readme

NoComment1105:
This module requires the Manage Webhook permission to avoid the internal `MessageCreateEvent` producing `KtorRequestExceptions` claiming there are missing permissions. This PR details this in the README.md of the module.

While I was there I updated gradle to 7.5


gdude2002:
o7

PR #175, opened 2022-08-20
Properly register dmPermission for global commands

Galarzaa90:
Since `GlobalChatInputCreateBuilder` inherits from `ChatInputCreateBuilder`, the `register` method for the parent class was always being called.

Since `GlobalChatInputCreateBuilder.register` was calling `ChatInputCreateBuilder.register` anyway, I think it's better to just add an if at the beginning to call `registerGlobalPermissions`

I added that `else` just to avoid calling `registerGuildPermissions` twice, since `registerGlobalPermissions` calls it too.

I tested it with `test-bot` but I did not commit those changes.


gdude2002:
Looks good to me, thanks!

PR #184, opened 2022-09-07
Fix subcommands with arguments with null parent command

Galarzaa90:
The parent command of subcommands with arguments were not being set properly.


gdude2002:
I can't tell what this fixes, but it doesn't appear to do what the PR description says it does? Wouldn't this always set "parent command" to the current command object?

Galarzaa90:
It is an extension function of `SlashCommand` to create a subcommand, so `this` refers to the parent command. Previously it was `parentCommand` so it was attempting to set it to the parent command's parent command (always being null).

If you check the same file, the extensions to create subcommands without arguments, already had `this` in there, it was just the ones with arguments that needed fixing.

Example
```kt
// This command's parentCommand is null, because it is a root command.
publicSlashCommand {
    name = "auction"
    description = "Show information about auctions."
    
    //This is the extension I'm modifying. Currently, it sets its parent command to the auction command's parent command.
   // What my change is doing is setting the parent command to `this` (auction command)
    publicSubCommand(::AuctionCharacterArguments) {
        name = "character"
        description = "Show auctions of a character."
        action { commandAuctionCharacter() }
    }

gdude2002:
Ah okay, I see. In that case, good spot, thanks!

PR #185, opened 2022-09-08
Add minLength and maxLength to StringConverter

Galarzaa90:
- Also changed visibility of minValue and maxValue properties in numeric converters

image


gdude2002:
Thanks!

PR #187, opened 2022-09-13
Set guildId to null when creating unsafe subcommands

Galarzaa90:
The corresponding functions for `ephemeralSubcommand` and `publicSubcommand` erase the `guildId` of the subcommand, but this was not being done for unsafeSubcommands.

Why didn't this raise an error before? Because until #184 was merged, all subcommands had `parentCommand = null`, so the validation never failed. 

I also missed one function when I created PR #184. So the parent command of unsafe subcommands with arguments was always null.


gdude2002:
Thanks!

PR #188, opened 2022-09-13
Added attachment option for slash commands

Galarzaa90:
I don't know if there were any plans to implement this differently, but I decided to try it out.

The only problem is that this converter does not work with chat commands, and it will always fail if used in a chat command. I don't know if I should do something about it.

image


kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/converters/impl/AttachmentConverter.kt
```diff
@@ -0,0 +1,46 @@
...
+    "attachment",
+
+    types = [ConverterType.OPTIONAL, ConverterType.SINGLE],
+)
+public class AttachmentConverter(
+    override var validator: Validator<Attachment> = null
+) : SingleConverter<Attachment>() {
+    override val signatureTypeString: String = "converters.attachment.signatureType"
+
+    override suspend fun parse(parser: StringParser?, context: CommandContext, named: String?): Boolean = false
```
gdude2002: This needs a more descriptive, translated error - thrown in a `DiscordRelayedException`. I think there's already a translation string you can use in there, but I'm not sure.

kord-extensions/src/main/resources/translations/kordex/strings.properties
```diff
@@ -68,6 +68,7 @@ commands.error.missingBotPermissions=I don't have the permissions I need to run
 commands.error.user=Unfortunately, **an error occurred** during command processing. Please let a staff member know.
 commands.error.user.sentry.message=Unfortunately, **an error occurred** during command processing. If you'd like to submit information on what you were doing when this error happened, please use the following command\: ```{0}feedback {1} <message>```
 commands.error.user.sentry.slash=Unfortunately, **an error occurred** during command processing. If you'd like to submit information on what you were doing when this error happened, please use the following command\: ```/feedback {0} <message>```
+converters.attachment.signatureType=attachment
```
gdude2002, line 71: This needs to be added to the English translations file as well.

gdude2002 (changes requested):

kord-extensions/src/main/resources/translations/kordex/strings.properties
Galarzaa90: Done, added to `strings_en_GB.properties`

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/converters/impl/AttachmentConverter.kt
Galarzaa90: I added a DiscordRelayedException, didn't find a translation string for it, added a new one.

I found a very similar case, for `ChoiceConverters`, when trying to use one in a chat command, it throws an IllegalStateException:
```kt
is ChoiceConverter<*> -> error("Choice converters may only be used with slash commands")
```
https://github.com/Kord-Extensions/kord-extensions/blob/develop/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/chat/ChatCommandParser.kt#L119

gdude2002: That's in the parser though, which isn't entirely ideal, yeah. I'll probably change how that works someday.

gdude2002:
Looking good - thanks!

PR #197, opened 2022-12-11
Fix kordex breaking itself because file names

sschr15:
All translations that I know about have been lowercased


gdude2002:
Your commits are unsigned, I'd appreciate you fixing that.

Also, it'd be worth looking into what caused this, imo - I definitely did not want these to be case sensitive.

sschr15:
I completely misunderstood the error, so this wouldn't fix things

PR #200, opened 2022-12-29
feat: more selector types

DeDiamondPro:
Adds support for user, role and channel selector types.


gdude2002:
Apologies for taking so long to get to this - would you mind checking out that conflict? I should be able to get this merged pretty soon.

DeDiamondPro:
> Apologies for taking so long to get to this - would you mind checking out that conflict? I should be able to get this merged pretty soon.

I'll fix the conflict when I have time tomorrow.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/components/_Functions.kt
```diff
@@ -97,53 +103,215 @@ public suspend fun <M : ModalForm> ComponentContainer.publicButton(
     return component
 }
 
-/** DSL function for creating an ephemeral select menu and adding it to the current [ComponentContainer]. **/
-public suspend fun ComponentContainer.ephemeralSelectMenu(
+/** DSL function for creating an ephemeral string select menu and adding it to the current [ComponentContainer]. **/
+public suspend fun ComponentContainer.ephemeralStringSelectMenu(
```
gdude2002, line 128: To ensure that bots don't suddenly stop working, I'd prefer we kept these functions around and deprecated them with a compiler error. This means that already-compiled bots won't break on it if the dependency in their environment is updated.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/components/menus/EphemeralSelectMenu.kt
```diff
@@ -19,19 +18,16 @@ import com.kotlindiscord.kord.extensions.utils.scheduling.Task
...
+import dev.kord.core.behavior.interaction.response.EphemeralMessageInteractionResponseBehavior
 import dev.kord.core.event.interaction.SelectMenuInteractionCreateEvent
-import dev.kord.rest.builder.message.create.InteractionResponseCreateBuilder
-
-public typealias InitialEphemeralSelectMenuResponseBuilder =
-    (suspend InteractionResponseCreateBuilder.(SelectMenuInteractionCreateEvent) -> Unit)?
 
-/** Class representing an ephemeral-only select (dropdown) menu. **/
-public open class EphemeralSelectMenu<M : ModalForm>(
+/** Interface for ephemeral select menus. **/
```
gdude2002: This is definitely still a class, not an interface. I don't think this doc comment needed changing at all, verboseness is good.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/components/menus/EphemeralSelectMenu.kt
```diff
@@ -19,19 +18,16 @@ import com.kotlindiscord.kord.extensions.utils.scheduling.Task
...
+public abstract class EphemeralSelectMenu<C, M : ModalForm>(
```
gdude2002, line 25: If `C` is meant to be some kind of context object, then you should give it a base type here.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/components/menus/PublicSelectMenu.kt
```diff
@@ -20,22 +19,19 @@ import dev.kord.common.annotation.KordUnsafe
...
+import dev.kord.core.behavior.interaction.response.PublicMessageInteractionResponseBehavior
 import dev.kord.core.event.interaction.SelectMenuInteractionCreateEvent
-import dev.kord.rest.builder.message.create.InteractionResponseCreateBuilder
-
-public typealias InitialPublicSelectMenuResponseBuilder =
-    (suspend InteractionResponseCreateBuilder.(SelectMenuInteractionCreateEvent) -> Unit)?
 
-/** Class representing a public-only select (dropdown) menu. **/
-public open class PublicSelectMenu<M : ModalForm>(
+/** Class for public select menus. **/
```
gdude2002: This really didn't need rewriting, bud. Let's not remove verboseness from existing comments.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/components/menus/PublicSelectMenu.kt
```diff
@@ -20,22 +19,19 @@ import dev.kord.common.annotation.KordUnsafe
...
+public abstract class PublicSelectMenu<C, M : ModalForm>(
```
gdude2002, line 26: Again, if `C` is a context type, you need to provide a base type for it. Don't leave generics untyped if at all possible.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/components/menus/string/StringSelectMenu.kt
```diff
@@ -0,0 +1,76 @@
...
+        // This is nullable like this due to a compiler bug: https://youtrack.jetbrains.com/issue/KT-51820
+        body: (suspend SelectOptionBuilder.() -> Unit)? = null,
+    ) {
+        val builder = SelectOptionBuilder(label, value)
+
+        if (body != null) {
+            body(builder)
+        }
+
+        if ((builder.description?.length ?: 0) > DESCRIPTION_MAX) {
```
gdude2002, line 45: These constants (e.g. `DESCRIPTION_MAX`) should be defined in the file they're used in. I think you forgot to move them over.

test-bot/src/main/kotlin/com/kotlindiscord/kord/extensions/testbot/extensions/SelectorTestExtension.kt
```diff
@@ -0,0 +1,169 @@
...
+                name = "public"
+                description = "Test public selectors."
+
+                action {
+                    respond {
+                        components {
+                            publicStringSelectMenu {
+                                option("hi", "1")
+                                option("hi hi", "2")
+                                maximumChoices = null
```
gdude2002, line 37: This seems like it should be a default, no? Why do you specify it here?

If it's not a default, it's probably time to make it so.

test-bot/src/main/kotlin/com/kotlindiscord/kord/extensions/testbot/extensions/SelectorTestExtension.kt
```diff
@@ -0,0 +1,169 @@
...
+                            }
+
+                            publicUserSelectMenu {
+                                maximumChoices = null
+
+                                action {
+                                    respond {
+                                        content = selected.joinToString("\n") {
+                                            runBlocking { it.asUser().username }
+                                        }
```
gdude2002: You shouldn't need `runBlocking` here or in any of the other action blocks. Isn't `joinToString` inline? Can you investigate why this is?

If there's nothing we can do about this, please store the result in a variable in the `action` block and make use of it in the `respond` block. Don't use `runBlocking`.

test-bot/src/main/kotlin/com/kotlindiscord/kord/extensions/testbot/extensions/SelectorTestExtension.kt
```diff
@@ -0,0 +1,169 @@
...
+                }
+            }
+
+            ephemeralSubCommand {
+                name = "ephemeral"
+                description = "Test ephemeral selectors."
+
+                action {
+                    respond {
+                        components {
```
gdude2002, line 95: See the comment I left on the other set of components.

gdude2002 (changes requested):
Overall, I think this is pretty good - just a few things to address. Good work!

test-bot/src/main/kotlin/com/kotlindiscord/kord/extensions/testbot/extensions/SelectorTestExtension.kt
DeDiamondPro: The default is a maximum of 1 choice, which is the same as it was before. Setting it to null makes max choices the same as the numbers of options you have (no maximum). doing `maximumChoices = 25` would serve the same purpose

gdude2002: Oh, hmm, okay, that's fair enough then.

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/components/menus/EphemeralSelectMenu.kt
DeDiamondPro: Since `C` needs to meet 2 conditions, it is defined in the where clause a few lines below (`where C : SelectMenuContext, C : EphemeralInteractionContext`)

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/components/menus/PublicSelectMenu.kt
DeDiamondPro: Since `C` needs to meet 2 conditions, it is defined in the where clause a few lines below (`where C : SelectMenuContext, C : PublicInteractionContext`)

gdude2002 (dismissed):
This is looking good - thanks for being patient. I think we can merge this once you've merged in the latest changes and verified it all works.

gdude2002:
Looks good, thanks for your work!

PR #202, opened 2023-01-03
Update to Kotlin 1.8, Gradle 7.6 and update all dependencies

NoComment1105:
Yeehaw. In the update of detekt, a lot of new errors were thrown by the detekt checks, that weren't previously thrown, therefore any errors that did come up, I disabled the checks instead of going through the project and making it comply to detekts rules.
I added the new detekt library rule plugin as some of the rules were already present in the normal file before they were deprecated and moved to the library rule plugin. Changes can be made if necessary


gdude2002:
Just a few notes.

1. Did you verify that the test bot works?
2. It seems you've disabled some Detekt rules that weren't new - is this intentional?
3. The Detekt yml files no longer end with a newline, please fix that

Otherwise, great work!

PR #203, opened 2023-01-28
Fix: fix mention for slash command groups

DeDiamondPro:
Move the slash command group check before the parentCommand check.
This fixes the mention of something like `/settings edit types` returning `/settings types`


gdude2002:
Hmm, I didn't think of that. Thanks!

PR #204, opened 2023-01-30
Update phishing and pluralkit extension to use nullable functions and…

NoComment1105:
… catch some more exceptions.

Previously, the phishing extension would relentlessly send exceptions for being unable to read URLs for various reasons, this is extremely frustrating since it's out of the users control. I've switch this to catch all exceptions, and log them at warn level instead, so they're still visible.

The PluralKit extension attempts to get channels using `asChannelOf` currently, resulting in class cast exceptions and null pointer exceptions. I have moves over to using `asChannelOfOrNull` to avoid the NPEs and surrounded initial attempts at using this function with a try/catch for `ClassCastExceptions`

### DIFF
--- a/extra-modules/extra-phishing/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/phishing/PhishingExtension.kt
+++ b/extra-modules/extra-phishing/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/phishing/PhishingExtension.kt
@@ -282,11 +282,7 @@ class PhishingExtension(private val settings: ExtPhishingBuilder) : Extension() 
         val domains: MutableSet<String> = mutableSetOf()
 
         for (match in settings.urlRegex.findAll(content)) {
-            val found = match.groups[1]?.value?.trim('/')
-
-            if (found == null) {
-                continue
-            }
+            val found = match.groups[1]?.value?.trim('/') ?: continue
 
             var domain = found
 
@@ -294,12 +290,12 @@ class PhishingExtension(private val settings: ExtPhishingBuilder) : Extension() 
                 domain = domain
                     .split("/", limit = 2)
                     .firstOrNull()
-                    ?.lowercase()
+                    ?.lowercase() ?: continue
             }
 
-            domain = domain?.filter { it.isLetterOrDigit() || it in "-+&@#%?=~_|!:,.;" }
+            domain = domain.filter { it.isLetterOrDigit() || it in "-+&@#%?=~_|!:,.;" }
 
-            if (domain in domainCache && domain != null) {
+            if (domain in domainCache) {
                 domains.add(domain)
             } else {
                 val result = followRedirects(match.value)
@@ -341,7 +337,7 @@ class PhishingExtension(private val settings: ExtPhishingBuilder) : Extension() 
 
             return url
         } catch (e: Exception) {
-            logger.debug(e) { e.message }
+            logger.warn(e) { url }
 
             return url
         }
--- a/extra-modules/extra-phishing/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/phishing/PhishingExtension.kt
+++ b/extra-modules/extra-phishing/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/phishing/PhishingExtension.kt
@@ -77,7 +77,7 @@ class PhishingExtension(private val settings: ExtPhishingBuilder) : Extension() 
             }
 
             action {
-                handleMessage(event.message)
+                handleMessage(event.message.asMessageOrNull())
             }
         }
 
@@ -92,7 +92,7 @@ class PhishingExtension(private val settings: ExtPhishingBuilder) : Extension() 
             }
 
             action {
-                handleMessage(event.message.asMessage())
+                handleMessage(event.message.asMessageOrNull())
             }
         }
 
@@ -140,7 +140,11 @@ class PhishingExtension(private val settings: ExtPhishingBuilder) : Extension() 
         }
     }
 
-    internal suspend fun handleMessage(message: Message) {
+    internal suspend fun handleMessage(message: Message?) {
+        if (message == null) {
+            return
+        }
+
         val matches = parseDomains(message.content)
 
         if (matches.isNotEmpty()) {
@@ -175,7 +179,7 @@ class PhishingExtension(private val settings: ExtPhishingBuilder) : Extension() 
                                 inline = true
 
                                 name = "Server"
-                                value = message.getGuild().name
+                                value = message.getGuildOrNull()?.name ?: "Unable to get guild"
                             }
                         }
                     }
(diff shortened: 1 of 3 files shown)


extra-modules/extra-phishing/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/phishing/PhishingExtension.kt
```diff
@@ -278,19 +282,24 @@ class PhishingExtension(private val settings: ExtPhishingBuilder) : Extension()
         val domains: MutableSet<String> = mutableSetOf()
 
         for (match in settings.urlRegex.findAll(content)) {
-            val found = match.groups[1]!!.value.trim('/')
```
gdude2002, line 281: I don't think it's possible for this to be missing (or if it is, the user supplied a custom regex and an error should be thrown)

extra-modules/extra-phishing/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/phishing/PhishingExtension.kt
```diff
@@ -278,19 +282,24 @@ class PhishingExtension(private val settings: ExtPhishingBuilder) : Extension()
...
+            if (found == null) {
+                continue
+            }
+
             var domain = found
 
             if ("/" in domain) {
                 domain = domain
                     .split("/", limit = 2)
-                    .first()
```
gdude2002, line 287: I don't think it's possible for this to be missing (or if it is, the user supplied a custom regex and an error should be thrown)

extra-modules/extra-phishing/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/phishing/PhishingExtension.kt
```diff
@@ -278,19 +282,24 @@ class PhishingExtension(private val settings: ExtPhishingBuilder) : Extension()
         val domains: MutableSet<String> = mutableSetOf()
 
         for (match in settings.urlRegex.findAll(content)) {
-            val found = match.groups[1]!!.value.trim('/')
+            val found = match.groups[1]?.value?.trim('/')
+
+            if (found == null) {
```
gdude2002: Should be an Elvis operator on the line above

extra-modules/extra-phishing/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/phishing/PhishingExtension.kt
```diff
@@ -278,19 +282,24 @@ class PhishingExtension(private val settings: ExtPhishingBuilder) : Extension()
...
-                    .first()
-                    .lowercase()
+                    .firstOrNull()
+                    ?.lowercase()
             }
 
-            domain = domain.filter { it.isLetterOrDigit() || it in "-+&@#%?=~_|!:,.;" }
+            domain = domain?.filter { it.isLetterOrDigit() || it in "-+&@#%?=~_|!:,.;" }
 
-            if (domain in domainCache) {
```
gdude2002: If the domain is null, we should have `continue`d or thrown earlier, given the logic you've written - it doesn't make sense to try to continue processing if there's no domain

extra-modules/extra-phishing/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/phishing/PhishingExtension.kt
```diff
@@ -331,12 +340,8 @@ class PhishingExtension(private val settings: ExtPhishingBuilder) : Extension()
...
 
             return url
-        } catch (e: HttpRequestTimeoutException) {
-            logger.warn { "$url -> Request timeout has expired." }
-
-            return url
-        } catch (e: URLParserException) {
-            logger.debug { "$url -> Failed to parse url" }
+        } catch (e: Exception) {
+            logger.debug(e) { e.message }
```
gdude2002: Your PR description states that this is a warning, not a debug message. Additionally, you've removed the URL from the log message - is that a good idea?

Furthermore, shouldn't logging the exception object by passing it directly to the logger call also output the exception message automatically?

extra-modules/extra-pluralkit/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/pluralkit/PKExtension.kt
```diff
@@ -149,15 +148,21 @@ class PKExtension : Extension() {
                     return@action
                 }
 
-                val channel = topChannelFor(event)!!.asChannelOf<TopGuildMessageChannel>()
+                val channel = try {
+                    topChannelFor(event)?.asChannelOfOrNull<TopGuildMessageChannel>()
+                } catch (e: ClassCastException) {
```
gdude2002, line 154: Isn't the point of `asChannelOfOrNull` to return `null` if the cast can't happen? I feel like this try-catch can never actually catch anything, though you should check what this function actually does before you take my word for it

gdude2002: It's also worth leaving a comment to point out that this works around Kord's lack of support for forum channels - otherwise it shouldn't really be possible for the cast to fail; all channels that can have webhooks are top guild message channels

extra-modules/extra-pluralkit/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/pluralkit/PKExtension.kt
```diff
@@ -230,15 +235,21 @@ class PKExtension : Extension() {
                     return@action
                 }
 
-                val channel = topChannelFor(event)!!.asChannelOf<TopGuildMessageChannel>()
+                val channel = try {
+                    topChannelFor(event)?.asChannelOfOrNull<TopGuildMessageChannel>()
```
gdude2002, line 241: Same thing applies to this block as in the comment on the previous one

extra-modules/extra-pluralkit/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/pluralkit/PKExtension.kt
```diff
@@ -266,7 +277,7 @@ class PKExtension : Extension() {
 
         event<MessageUpdateEvent> {
             action {
-                val guild = event.channel.asChannelOfOrNull<GuildChannel>()?.getGuild()
+                val guild = event.channel.asChannelOfOrNull<GuildChannel>()?.getGuildOrNull()
```
gdude2002, line 282: This should never return null unless the guild isn't in the cache - have you actually run into this one?

extra-modules/extra-pluralkit/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/pluralkit/PKExtension.kt
```diff
@@ -297,15 +308,21 @@ class PKExtension : Extension() {
                     return@action
                 }
 
-                val channel = topChannelFor(event)!!.asChannelOf<TopGuildMessageChannel>()
+                val channel = try {
```
gdude2002, line 314: Usual comments for this block apply.

gdude2002 (changes requested):
Appreciate the PR, but I think there are some issues. Please take a look at the comments and see what you think.

extra-modules/extra-pluralkit/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/pluralkit/PKExtension.kt
NoComment1105: > Isn't the point of `asChannelOfOrNull` to return `null` if the cast can't happen? I feel like this try-catch can never actually catch anything, though you should check what this function actually does before you take my word for it

Yes, however within the KDoc of the function it details that is will still throw if the channel is not of the specified type. I've not experimented enough to determine whether this is a copy-paste issue in the docs or factually correct.

extra-modules/extra-pluralkit/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/extra/pluralkit/PKExtension.kt
NoComment1105: I've not run into it yet, though I'd personally prefer to cover this base in the event of it cropping up in the future

gdude2002:
This is more or less good - thanks for the contribution!

PR #205, opened 2023-02-16
Address Deprecations and unused suppressions/optins

NoComment1105:
Addresses the deprecations of `getGuild` for `getGuildOrNull` methods and removes suppressions and OptIn annotations that where unused


gdude2002:
Thanks!

PR #211, opened 2023-04-05
Fix breaking changes from Kord Kotlin/JS support

lukellmann:
see https://github.com/kordlib/kord/pull/775


kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/utils/_Permissions.kt
```diff
@@ -57,6 +57,8 @@ public fun Permission.toTranslationKey(): String? = when (this) {
 
     Permission.UseExternalStickers -> "permission.useExternalStickers"
     Permission.UseEmbeddedActivities -> "permission.useEmbeddedActivities"
+    Permission.ViewCreatorMonetizationAnalytics -> TODO()
+    Permission.UseSoundboard -> TODO()
```
lukellmann: it's a draft pr because i don't know what belongs here

NoComment1105: Looks to just be the translation key, i would assume just
```suggestion
    Permission.ViewCreatorMonetizationAnalytics -> "permission.viewCreatorMonetizationAnalytics"
    Permission.UseSoundboard -> permission.useSoundboard
```
then add to the translation files those lines. Just add it to the root strings.properties i believe

gdude2002: That one and the UK one yeah

lukellmann: done

gdude2002 (approved):

gdude2002:
Thanks for that!

PR #212, opened 2023-04-12
Fix compilation error caused by source-incompatible change from kord#703 (#212

lukellmann:
see https://github.com/kordlib/kord/pull/703


lukellmann:
I hope that it's ok to change types of public properties and constructors in these places.

gdude2002:
Looks good to me, thanks!

PR #213, opened 2023-04-27
Update dependencies and update for kord 0.9.0

NoComment1105:


gdude2002:
Thanks - not sure why Licenser broke there, but I'll figure it out.

PR #222, opened 2023-07-21
SlashCommand: Allow having groups and subcommands on the same level

Leo40Git:
this is legal now

Note that I haven't reconfigured the maximum subcommand/group limit (it still applies to subcommands and groups individually), and I can't actually find mention of this limit in the official documentation.

### DIFF
--- a/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/slash/_Functions.kt
+++ b/kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/slash/_Functions.kt
@@ -32,10 +32,6 @@ public suspend fun SlashCommand<*, *, *>.group(name: String, body: suspend Slash
         error("Command groups may not be nested inside subcommands.")
     }
 
-    if (subCommands.isNotEmpty()) {
-        error("Commands may only contain subcommands or command groups, not both.")
-    }
-
     if (groups.size >= SUBCOMMAND_AND_GROUP_LIMIT) {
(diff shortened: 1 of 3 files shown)


kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/slash/_Functions.kt
```diff
@@ -32,10 +32,6 @@ public suspend fun SlashCommand<*, *, *>.group(name: String, body: suspend Slash
         error("Command groups may not be nested inside subcommands.")
     }
 
-    if (subCommands.isNotEmpty()) {
```
gdude2002, line 35: Isn't there also a set of subcommand functions that check whether there's a group? Please double-check that if you can, I don't have time atm

gdude2002 (changes requested):

kord-extensions/src/main/kotlin/com/kotlindiscord/kord/extensions/commands/application/slash/_Functions.kt
Leo40Git: None of the `publicSubCommand` and `ephemeralSubCommand` functions in this file had a similar check (so theoretically, bots _could_ have already been using mixed groups/subcommands if they registered in the correct order).

gdude2002: Oh, balls, okay. Well, thanks!

gdude2002:
Thanks for that, good shout!

PR #238, opened 2023-09-14
mongo adapter: Allow reading of old INT64 Snowflake from databases

NoComment1105:
Without this, any Snowflake entry in a database record that was created using KMongo, which serialized Snowflakes as INT64, will throw an error. This allows proper decoding of those INT64 entries, alongside the STRING entries


gdude2002 (changes requested):
Doesn't build - presumably a missing import?

NoComment1105:
Whoops, will fix later

gdude2002 (approved):

gdude2002:
Cheers

PR #239, opened 2023-09-15
Set guildId to null when registering unsafeSubCommand in group

Galarzaa90:
This should fix the error message shown when using a `defaultGuild` in the bot's configuration and registering a `unsafeSubcommand` under a `group`


gdude2002:
Thanks!

PR #267, opened 2023-12-14
Fix conversion commands

sschr15:
Local testing showed that it does work as intended now


gdude2002:
It doesn't build, so I'm not sure I believe that

gdude2002:
After some chatting, realised this will pass in prod.

gdude2002:
Thank

PR #279, opened 2024-02-28
Update to Linkie patch 121

sschr15:
Also removes the temporary fix for MCP and adds a proper error message in one of the possible error conditions


extra-modules/extra-mappings/src/main/resources/translations/kordex/mappings.properties
```diff
@@ -16,3 +16,4 @@ command.mapping.namespace.description=Configure which namespaces can be used in
 command.mapping.namespace.selectmenu=Select the namespaces for conversion commands
 command.mapping.namespace.selectmenu.cleared=Namespaces cleared
 command.mapping.namespace.selectmenu.updated=**Namespaces updated:** {0}
+command.mapping.nodefault=**Error:** no default version could be loaded for namespace {0}
```
gdude2002: Shouldn't this tell users how to solve the problem? As in, supply a version?

gdude2002 (changes requested):
Looks good, just a small change.

gdude2002 (approved):

gdude2002:
Thanks!

PR #301, opened 2024-06-03
Fix Exception when running with KSP2

DRSchlaubi:
Seemingly Enum constants are now KSClassDeclarations rather than KSTypes

### DIFF
--- a/annotation-processor/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/annotations/converters/ConverterAnnotationArgs.kt
+++ b/annotation-processor/src/main/kotlin/com/kotlindiscord/kord/extensions/modules/annotations/converters/ConverterAnnotationArgs.kt
@@ -9,7 +9,7 @@
 package com.kotlindiscord.kord.extensions.modules.annotations.converters
 
 import com.google.devtools.ksp.symbol.KSAnnotation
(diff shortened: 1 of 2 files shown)


gdude2002:
The branch you've targeted does not support K2 - did you mean to target the K2 branch?

DRSchlaubi:
> The branch you've targeted does not support K2 - did you mean to target the K2 branch?

I specifically said KSP2 not K2, the project seems to work fine on KSP2 when not using k2

gdude2002:
Is there a reason to rush the usage of KSP 2 before we can even use K2?

DRSchlaubi:
> Is there a reason to rush the usage of KSP 2 before we can even use K2?

I want to upgrade my project to KSP2 and this prevents me from doing so haha

Also it doesn't seem to cause issues with k1 users, so why not upgrade to KSP2 while we can, seems to have no downside

gdude2002:
Did you test the converter system via the test bot?

DRSchlaubi:
> Did you test the converter system via the test bot?

I did not, I assumed if KSP would cause issues they would be compile-time issues, not runtime issues

gdude2002 (approved):
Approving after test confirmation on Discord.

gdude2002:
Thanks!

PR #306, opened 2024-07-10
Use mappings for EventHandler sentry context

Galarzaa90:
Giving these the same treatment as #295 

image

Using a mapping, and merging extension name into it to match the structure used in commands.


gdude2002 (approved):

gdude2002:
Thanks!

PR #360, opened 2025-01-15
Check permissions before running phishing actions

NoComment1105:
Previously an error would be thrown if permissions were missing for a detection action in the phishing module. This PR adds some checks in to ensure the permissions are present in the bot before running those actions. If the permissions aren't present then it will send a warning to the console logs and insert a field to the logging embed that states permissions are missing.

I do have a question, should the checks on ban and kick be seperate to the delete check so that if the bot has one but not the other permission, one of the actions still gets run?

I haven't PR'd the fields to the translations repo yet, wanted to ensure the code was acceptable first.


gdude2002:
> I do have a question, should the checks on ban and kick be seperate to the delete check so that if the bot has one but not the other permission, one of the actions still gets run?

Yeah, I think so.

> I haven't PR'd the fields to the translations repo yet...

It does look like you'll need to do translations for the action type as well? I don't recall whether I already did that.

---

Otherwise, looks fine!

NoComment1105:
Changed made and PR to translations made 👍🏼

gdude2002 (approved):

gdude2002:
Thanks!

PR #365, opened 2025-03-19
support MemberLeaveEvent in memberFor() helper

UpcraftLP:
Since `MemberLeaveEvent` only extends the root `Event` type, it needs special handling in `memberFor()`.

`userFor()` and `guildFor()` already handle it, so I believe this was a simple oversight.


gdude2002 (approved):
👍

gdude2002:
Thanks!

Kord doesn't make it easy to keep up with things like this, esp when events are added or modified... I wonder if there's any better approaches we could take?

PR #367, opened 2025-04-05
WIP: Lazy Pagination

### LINKED ISSUE #90: Lazy loading on Paginators

NinekoTheCat:
Fixes: #90 by adding a new type of paginator

## Changes
- Added LazyPages and tests for it.
- Added `Pages` as an interface to replace the static `Pages` class input into Paginators.
- Added custom rule to avoid prefixing interfaces with I
## TODOS:
- [ ] Modify paginators to actually recognize and handle endless pages
- [ ] Add more options for Pages types to support previous/next hasPrevious/hasNext


gdude2002:
I understand that this is a draft, but just so you're aware, the project convention is to avoid `IName` interfaces, instead preferring a `Name` interface with a `DefaultName` implementation where a default implementation makes sense.

In this case, I believe that means a `Pages` interface, with `DefaultPages` and `LazyPages` implementations, right?

Otherwise, looking good so far.

NinekoTheCat:
> I understand that this is a draft, but just so you're aware, the project convention is to avoid `IName` interfaces, instead preferring a `Name` interface with a `DefaultName` implementation where a default implementation makes sense.
> 
> In this case, I believe that means a `Pages` interface, with `DefaultPages` and `LazyPages` implementations, right?
> 
> Otherwise, looking good so far.

I renamed them to the proper names

gdude2002:
Your last commit was last month, but your PR hasn't been marked ready to review - how's this going? Do you need any help?

Also, just so you're aware, you're signing off your commits twice. There also seems to be a merge conflict?

gdude2002:
This PR hasn't been touched since my last comment a few months ago, so I'm going to consider it abandoned for now. Please do open another one if/when you get back to this feature - it would be incredibly useful!

PR #370, opened 2025-08-10
replace deprecated method call

UpcraftLP:


gdude2002:
This was a little unnecessary, since I will be replacing this entire plugin (more or less) later, but thanks all the same.

PR #372, opened 2025-08-10
security hardening for GH actions

UpcraftLP:
pin GH actions workflows to specific commit IDs


gdude2002 (approved):
I was a little confused at first, but actually, I guess this does make sense! Thanks.

PR #373, opened 2025-08-10
install nodeJS and pnpm on CI instead of gradle

UpcraftLP:
Making the NodeJS and PNPM installations managed by GH actions instead of gradle allows for caching of node dependencies.


gdude2002 (approved):
Cheers!

PR #385, opened 2025-08-24
Added an improved error message when a scheduled task fails.

Taubsie:


gdude2002:
Seems like a pretty safe and positive change. Cheers!

PR #405, opened 2026-01-25
Fixed the Data Collector sometimes failing on Kord versions with changed Intents.

Taubsie:


gdude2002:
Hey, thanks for the PR!

Unfortunately, this isn't a practical approach for a few reasons, the main one being that it makes it far too easy for me to miss translations when new intents are added, and translations that I need to remove when intents are removed.

This is essentially a one-person dev team, so I wouldn't be comfortable creating more maintenance work for myself.

PR #413, opened 2026-07-25
Generalize modal components and add support for text displays

Tmpod:
Previously, KordEx's `ModalForm` API only supported interactive `Widgets`, which return user input. In the current Discord API though, you can have [Text Displays](https://docs.discord.com/developers/components/reference#text-display) on modals too, at the root level.

This commit generalizes the root modal API a bit so it can support these text display components, while keeping the existing `Widget` API pretty much intact.

### DIFF
--- a/kord-extensions/src/main/kotlin/dev/kordex/core/components/forms/_Grid.kt
+++ b/kord-extensions/src/main/kotlin/dev/kordex/core/components/forms/_Grid.kt
@@ -10,10 +10,9 @@
 
 package dev.kordex.core.components.forms
 
-import dev.kordex.core.components.forms.widgets.Widget
 import dev.kordex.core.utils.map
 
-public typealias WidgetGrid = Array<MutableList<Widget<*>?>>
+public typealias WidgetGrid = Array<MutableList<Widget?>>
 
 /** A coordinate pair, represented by a `Pair(row, column)`. **/
 public typealias CoordinatePair = Pair<Int, Int>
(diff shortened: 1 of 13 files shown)


kord-extensions/src/main/kotlin/dev/kordex/core/components/forms/FormComponent.kt
```diff
@@ -0,0 +1,34 @@
...
+
+package dev.kordex.core.components.forms
+
+import dev.kord.rest.builder.interaction.ModalBuilder
+import java.util.*
+
+/**
+ * A component usable inside modal forms.
+ */
+public abstract class FormComponent {
```
gdude2002: you've abstracted this the other way around from project convenctions! we would normally do, eg, `Widget` and `InteractibleWidget`, this PR essentially gives users a `Widget` and `UninteractibleWidgetParent`, which is a bit confusing! --luna

kord-extensions/src/main/kotlin/dev/kordex/core/components/forms/_Grid.kt
```diff
@@ -10,10 +10,9 @@
```
gdude2002, line 10: in line with my other comment, the renames in this file should be reverted! --luna

gdude2002 (changes requested):
thanks for your contribution! just a couple small nitpicks c: 

and if you could bump the minor version as well, we'd appreciate that! c: -- luna

kord-extensions/src/main/kotlin/dev/kordex/core/components/forms/_Grid.kt
```diff
@@ -153,7 +152,7 @@ public fun CoordinatePair.toString(): String =
 public fun WidgetGrid.toString(): String = buildString {
 	val grid = this@toString
 
-	appendLine("WidgetGrid@${hashCode()} -> ")
+	appendLine("FormComponentGrid@${hashCode()} -> ")
```
gdude2002: found a forgor! --luna

gdude2002 (changes requested):
just quickly found a missing rename - otherwise, i'm happy! --luna

kord-extensions/src/main/kotlin/dev/kordex/core/components/forms/_Grid.kt
Tmpod: oopsie, fixed!

